PR #60, opened 2014-09-18
[all] allow skipping kernel config checker

sledges:
DO NOT MERGE

we might want to rework the logic of this a bit.


sledges:
Superseded by https://github.com/mer-hybris/droid-hal-device/pull/214

PR #75, opened 2014-11-25
New device patterns for OnePlus One (bacon)

RussellHaley:
Did my best to follow instructions from Umeaboy. Thanks


mlehtima:
This seems to be obsolete as bacon support is in separate repos after refactoring.

PR #78, opened 2015-01-11
[mako] Enable Bluetooth HCI SMD during boot.

zanac:
This needs CONFIG_BT_HCISMD=y in the kernel.  For some reason, writing to the sysfs files using tmpfiles.d at boot doesn't work - you need to wait some amount of time before writing to the file.

### DIFF
--- a/device-lge-mako-configs/lib/systemd/system/bluetooth.service.wants/hcismd.service
+++ b/device-lge-mako-configs/lib/systemd/system/bluetooth.service.wants/hcismd.service
@@ -1,0 +1,1 @@
+../hcismd.service
--- a/device-lge-mako-configs/lib/systemd/system/hcismd.service
+++ b/device-lge-mako-configs/lib/systemd/system/hcismd.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Enable Bluetooth HCI over SMD
+DefaultDependencies=false
+After=local-fs.target
+
+[Service]
+Type=oneshot
+ExecStart=/sbin/hcismd-enable.sh
+RemainAfterExit=yes
+
+[Install]
+WantedBy=bluetooth.service
--- a/device-lge-mako-configs/sbin/hcismd-enable.sh
+++ b/device-lge-mako-configs/sbin/hcismd-enable.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+# Maximum number of attempts to enable hcismd to try to get
+# hci0 to come online.  Writing to sysfs too early seems to
+# not work, so we loop.
+MAXTRIES=15
+
+seq 1 $MAXTRIES | while read i ; do
+    echo 1 > /sys/module/hci_smd/parameters/hcismd_set
+    if [ -e /sys/class/bluetooth/hci0 ] ; then
+        # found hci0, exit successfully
+        exit 0
+    fi
+    sleep 1
+done
+# must have gotten through all our retries, fail
+exit 1


sledges:
Notes from @hmallat: at some point I was experimenting [on Jolla, similar HW to mako] with having a startup script that waits until a hci interface exists, but that needed a bit of C (at least I don't know how I could do it with only shell scripting)
but possibly e.g. waiting until rfkill switch for hciX appears could be done instead of a blind sleep

zanac:
Yeah I'm definitely open to alternatives on how to do this, but writing to this sysfs file is what's going to *cause* hciX and the rfkill switch to appear.  So at least I can't wait on those.

hmallat:
Ok, it might be necessary to do some C coding then; you can open a socket(AF_BLUETOOTH, SOCK_RAW, BTPROTO_HCI), set it up to filter for internal stack events, and wait until you see a device registration event from the kernel. It's a bit of work, but on the other hand there's not really any guarantee that 10 seconds of waiting would be enough.

zanac:
Sure, I'll try that out.  I definitely agree that the sleep is an unpleasant way to do things.

I'm just not familiar enough with all of the pieces involved here - it's still my impression that none of the HCI stuff will even start existing until after the write to sysfs, but I'd be happy to be proven wrong.

zanac:
Hmm, one simple thing to look for is the presence of /sys/class/bluetooth/hci0.  That starts existing once the write to hcismd_set does something.  So a shell script would be sufficient to watch for that.

If that's ok, where should this live?  I'm thinking /usr/sbin?  And I imagine we'd want an upper bound for how long to retry for.. maybe like a minute?

sledges:
but the write to hcismd_set still has to be delayed, if I understand correctly?

zanac:
Looks like you can write it as many times as you want, but it only starts having an effect some time after boot.  So writing it too soon (during tmpfiles) means it does nothing.  Once the effect has occurred (i.e. hci0 starts existing), you can continue writing to it with no bad effects.  So I'd propose a shell script that writes it in a loop (with some kind of short sleep, giving up after some maximum number of loops) that writes it and checks for the existence of hci0.

sledges:
It can live here for example: https://github.com/mer-hybris/droid-hal-device/commit/34ab62c4312039b3fd3648bb348a830ff808e39a

Pick an upper bound empirically, we can always adjust it later

zanac:
I've tried rebooting a number of times and the hci0 device appears 7 seconds after systemd starts the hcismd-enable.sh script.  So I've given it 15 seconds for a safe margin of error.

zanac:
Hm.. bluetoothd is smart enough to cope with hci0 appearing either before or after it starts.  And hcismd can succeed whether or not bluetoothd is running or will ever run.  So I don't think it's appropriate to add an After=.

IMO it's sufficient to just use bluetooth.service.wants, because we just want them to happen more or less together, with no requirements on ordering.

But I've also noticed the shell script I wrote can't fail - I should probably make it so the systemd unit fails if for some reason hci0 didn't appear after all of the loops.  And I'll tidy up the commits too.

sledges:
Then add After=local-fs.target , it's uncommon to have .service files without After or Before.

Also, if symlinking under bluetooth.service.wants draws them together, that's just better (but I didn't know it would guarantee such timings)

zanac:
Ok, thanks for the patience - I think we've finally got something we can all be happy with.  :)

sledges:
it worked! and paired with another phone immediately and transfered a file :D job done! and well done @zanac! 
before i merge - pls put it under /lib instead of /etc :) (or tell my why did you prefer /etc, as we (or well, at least other porters more experts than me in systemd or just following the first one) go for /lib?)

zanac:
I was just looking at the hammerhead configs, and was just plain accustomed to making my own local systemd customizations in /etc.  No strong reason.  So they're in /lib now.

I don't know if you've been able to test it, but I haven't been able to make a phone call over bluetooth yet.  To my untrained eyes it looks like some kind of problem with the pulseaudio configuration, but I suspect that's a pull request for another day.  :)  At least this is already a major improvement.

sledges:
I haven't been able to pair it with me Motorola BT Keyboard either, so surely some digging ahead:)

Sorry for not double-checking last time, but you forgot this bit (sic): https://github.com/mer-hybris/droid-hal-device/commit/6e1888483593054f216122c44ef76c28ceea5914

zanac:
Hm.. is something missing?  It's 755 in my working copy and when I reclone it from github.

hmallat:
sledges: re keyboards, I had to bring in some HID fixes both to Jolla phone and hammerhead. I forget the exact details as it's been a while, but at least some HID qualification tests had trouble passing without the fixes. See https://github.com/mer-hybris/android_kernel_lge_hammerhead/pull/5

sledges:
ok, cloned, executable bit set, merging \o/

PR #79, opened 2015-01-14
[hammerhead][wlan] Modularising will ensure correct MAC. Contributes to ...

sledges:
...NEMO#731

Built-in module gets WLAN chip before persistent partition is mounted,
which contains its HW MAC address.

Hence we need to modularise WLAN


sledges:
@saukko, yes: https://github.com/mer-hybris/droid-hal-device/blob/hybris-10.1/makefstab#L107

spiiroin:
While "nb#" looks familiar, is it correct? I recall that "NEMO#" at least used to be the thing.

sledges:
@spiiroin, you are right! Fixed (as per http://www.merproject.org/meetings/mer-meeting/2011/mer-meeting.2011-12-13-11.02.log.html )

sledges:
@saukko gave his "LGTM"

PR #86, opened 2015-04-01
Enable usb moded

sledges:
Merge with https://github.com/mer-hybris/droid-hal-configs/pull/13 counterpart, as SFE community is still using monolithic dhd packaging. Split modular packaging is WIP


siteshwar:
LGTM

PR #88, opened 2015-04-09
[tools] Add option for building device specific droid-hal-tools

kjokinie:
Sometimes hardware vendors add custom features to the default android
image creation tools in their BSPs, or the BSP requires a specific
version of those tools. Added option "enable_device_specific_tools"
that can be set 1 in device specific spec file to build a custom
droid-hal-tools specific to that device BSP instead of using the default
droid-hal-tools.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -10,6 +10,7 @@
 # device_target_cpu: Used for native builds. Normally the nested droid build env will do an old-fashioned cross-compile and produce non-x86 binaries (default armv7hl). This can be set to tell OBS what arch the binaries are. Eg Android for Intel arch must set this.
 # device_variant: for AOSP this is used as the TARGET_BUILD_VARIANT for lunch
 # makefstab_skip_entries: Allow entries into the makefstab unit creation to be skipped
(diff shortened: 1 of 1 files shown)


kjokinie:
This requires first a change in the hybris-boot to actually build the tools in the BSP: https://github.com/mer-hybris/hybris-boot/pull/47

kjokinie:
Discussed with sage and seems this does not fly as it is now.

The problem is that these tools get built for host x86 only in this setup. And currently droid-hal-tools can be run on target as well. So the packaging needs not to obsolet droid-hal-tools but instead  package a different package called *-host-tools, and we have the binaries postfixed with *-host, so they don't interfere with the binaries from droid-hal-tools.

kjokinie:
Version 3 or something around that:

* Does not obsole droid-hal-tools anymore

* Rename the tools to "host tools" as they don't really run on target

kjokinie:
Version 4:

* Removed use of ENABLE_HYBRIS_TOOLS, as the tools will now get always built

sledges:
LGTM

PR #89, opened 2015-04-14
Revert "[tools] Add option for device specific image creation tools"

kjokinie:
This reverts commit f6bf842884bd5dfa458dd8b2b868337fc2c78d6f.

If device specific tools are needed, one should just modify a
device specific droid-hal-tools package for their adaptation.
It is much quicker and works better with legacy functionality.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -101,6 +101,22 @@ Device specific droid headers for %{rpm_device}.
 Needed by libhybris
 
 ################
+%package tools
+Provides: droid-hal-tools
+Group:	Tools
+Summary: Some tools from android specific for %{device}
+License: ASL 2.0
+BuildRequires:  pkgconfig(zlib)
+
+Obsoletes: droid-hal-tools <= 0.0.2
+
+%description tools
+This package contains some tools that some hardware adaptations need for
+handling the android style content. This package is intended to be such
+that it can be installed on the device and is not meant to be used in host/sdk
+environment.
+
+################
 %package kernel
 Provides: droid-hal-kernel
 Group:	System
@@ -190,6 +206,19 @@ echo Verifying kernel config
 hybris/mer-kernel-check/mer_verify_kernel_config \
     %{android_root}/out/target/product/%{device}/obj/*/.config
 
+# Check if local overrides for tool mk files exist, otherwise dhd defaults are used
+if test -f rpm/helpers/mkbootimg.mk; then
+    export MKBOOTIMG_MK=$(pwd)/rpm/helpers/mkbootimg.mk
+fi
+
+if test -f rpm/helpers/simg2img.mk; then
+    export SIMG2IMG_MK=$(pwd)/rpm/helpers/simg2img.mk
+fi
+
+if test -f rpm/helpers/img2simg.mk; then
+    export IMG2SIMG_MK=$(pwd)/rpm/helpers/img2simg.mk
+fi
+
 echo Building local tools
 ANDROID_ROOT=$(readlink -e %{android_root})
 (cd %{dhd_path}/helpers; make ANDROID_ROOT=$ANDROID_ROOT)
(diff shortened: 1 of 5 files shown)


kjokinie:
Version 2 of the PR:

Now includes a new patch the builds the tools same way as the generic droid-hal-tools OBS package does. Please check the patch description for more details.

droid-hal-device.inc
```diff
@@ -191,6 +197,29 @@ echo Running droid build in HABUILD_SDK
 ubu-chroot -r /srv/mer/sdks/ubu "cd %android_root; source build/envsetup.sh; lunch %{device}%{?device_variant}; rm -f .repo/local_manifests/roomservice.xml; make %{?_smp_mflags} %{?hadk_make_target}%{!?hadk_make_target:hybris-hal}"
 %endif
 
+# tools package: look first for device specific mk file, then use dhd's
+rm -f %{android_root}/system/core/mkbootimg/mkbootimg \
+      %{android_root}/system/core/libsparse/img2simg \
+      %{android_root}/system/core/libsparse/simg2img
+
+if test -f rpm/helpers/mkbootimg.mk; then
+    make -f `pwd`/rpm/helpers/mkbootimg.mk -C %{android_root}/system/core/mkbootimg/
```
sledges: `pwd` here and below, should not be needed, as build dir will always be $ANDROID_ROOT (where rpm/ resides)

droid-hal-device.inc
```diff
@@ -191,6 +197,29 @@ echo Running droid build in HABUILD_SDK
...
+else
+    make -f `pwd`/rpm/dhd/helpers/mkbootimg.mk -C %{android_root}/system/core/mkbootimg/
```
sledges: replace rpm/dhd with %{dhd_path}, here and below

droid-hal-device.inc
```diff
@@ -102,6 +101,22 @@ Device specific droid headers for %{rpm_device}.
 Needed by libhybris
 
 ################
+%package tools
+Provides: droid-hal-tools
```
saukko: We do not need this as hw adaptation repo can make its own decision which package to provide and after this we do not need to have droid-hal-tools at all anymore.

kjokinie:
Thanks, good comments. I'll fix and update PR.

droid-hal-device.inc
kjokinie: Hmm.. seems when using the -C option for make makes the current pwd change for duration of the make run, and then the line does not work without the pwd. Tried with mb2 building.

kjokinie: In other words, you need to have absolute path for the .mk file there.

kjokinie:
Version 3:

* Removed `pwd` from test lines, as it's not needed there

* Replaced depracated `...` with $(...) calls

* Use %{dhd_path} istead of rpm/dhd/

droid-hal-device.inc
```diff
@@ -191,6 +197,29 @@ echo Running droid build in HABUILD_SDK
 ubu-chroot -r /srv/mer/sdks/ubu "cd %android_root; source build/envsetup.sh; lunch %{device}%{?device_variant}; rm -f .repo/local_manifests/roomservice.xml; make %{?_smp_mflags} %{?hadk_make_target}%{!?hadk_make_target:hybris-hal}"
 %endif
 
+# tools package: look first for device specific mk file, then use dhd's
+rm -f %{android_root}/system/core/mkbootimg/mkbootimg \
+      %{android_root}/system/core/libsparse/img2simg \
+      %{android_root}/system/core/libsparse/simg2img
+
+if test -f rpm/helpers/mkbootimg.mk; then
+    make -f $(pwd)/rpm/helpers/mkbootimg.mk -C %{android_root}/system/core/mkbootimg/
```
sledges: We should keep it consistent with current "make" approach used here: https://github.com/mer-hybris/droid-hal-device/blob/f6bf842884bd5dfa458dd8b2b868337fc2c78d6f/droid-hal-device.inc#L205
It also needs testing whether this line (make -f $(pwd)/rpm... , or after fixing the bit above: (cd rpm/; make ...) ) will build on OBS wrt to correct paths. @lbt, I think your glance is enough, instead of full test build.

kjokinie:
Version 4:

* Moved the tools build into the normal /helpers/makefile

* Overrides checked at dhd.inc and passed over via <TOOLNAME>_MK environment variables to helpers makefile build

sledges:
LGTM

PR #90, opened 2015-04-15
Add possibility to add post scripts via rpm macro.

saukko:
Some adaptations might need to have their own post scripts that are device specific, lets allow this by adding macro for it.

Signed-off-by: Marko Saukko <marko.saukko@jolla.com>


sledges:
LGTM

PR #91, opened 2015-04-20
[tools] Fix img2simg build. Fixes JB#28015

kjokinie:
The output_file.c source uses too smart macro magics for it's own good,
which doesn't play well with C99. Disabled -std=c99 option for
img2simg to make it build without errors. Same added for simg2img
as it builds against same object files.


sledges:
LGTM

PR #94, opened 2015-06-02
[dhd] fix typo, which would prevent a potential breakage

sledges:


sledges:
@lbt sat next approved it :P

PR #95, opened 2015-06-11
[packaging] avoid macro name clash across projects during SR

sledges:


sledges:
Hm not happy with this, not elegant  enough;P

PR #96, opened 2015-06-18
do timestamped releases only for local HADK builds

sledges:
HADK dhd builds that still use RPM upload to OBS, should still have rel_date
timestamp as release, and therefore should define local_hadk_build_project
macro (currently provisioned under /usr/lib/rpm/macros.d/)

All OBS builds (including QA checkers) should not have local_hadk_build_project
macro defined.


iamer:
LGTM. In general I still think that a spec file that changes everytime you parse it is not such a good idea :)

sledges:
:+1:

PR #99, opened 2015-07-09
Handle _build_flavour as string

mkosola:


droid-hal-device.inc
```diff
@@ -60,7 +60,7 @@
 %define _obs_build_count %(echo %{release} | awk -F . '{if (NF >= 3) print $3; else print $1 }')
 %define _obs_commit_count %(echo %{release} | awk -F . '{if (NF >= 2) print $2; else print $1 }')
 
-%if %{_build_flavour} == release
+%if "%{_build_flavour}" == release
```
sledges: shall we say "release" too for good measure?

sledges:
LGTM

PR #102, opened 2015-08-12
[multirom] add the initramfs to the kernel rpm. contributes towards a…

alinelena:
… working zypper dup with multirom

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -414,6 +414,8 @@ touch $RPM_BUILD_ROOT/%{_libdir}/droid/droid-user-remove.sh.installed
 # /boot and modules to /lib as normal
 KERNEL_RELEASE=$(cat out/target/product/%{device}/*/*/include/config/kernel.release)
 cp out/target/product/%{device}/kernel $RPM_BUILD_ROOT/boot/kernel-$KERNEL_RELEASE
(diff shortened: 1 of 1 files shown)


sledges:
@saukko, currently MultiROM devices don't have OTA because kernel cannot be flashed. MultiROM keeps kernel and initrd images on plain filesystem. We could split boot.img on device, but that's too costly, so we'd better ship initramfs together with kernel rpm, agree?

saukko:
@sledges I think we should add macro has_multirom or something so that it would not change anything on existing systems.

sledges:
then it's double work. MultiROM is distinguished by cmdline. Otherwise image build is exactly the same, and we don't want to end up building two images

sledges:
delete_files is not in dhd, only in dh-configs for now; can we compromise on has_native_bootloader_only and put it on non-multirom devices, in this way the amount of modifications across all devices will be minimal (most ported devices don't mind having initrd shipped, and also they might get multirom eventually as well)

sledges:
We forgot that ```have_custom_img_boot``` can be used for providing own whole package for e.g. .img type of adaptations as opposed to .zip, where partitioning/flashing is involved -- in that case it's certainly not a multirom device, so merging this PR is safe and not too cluttery.
Don't forget, we will also need adding -kernel package to patterns: https://github.com/mer-hybris/droid-config-hammerhead/blob/master/patterns/jolla-hw-adaptation-hammerhead.yaml#L13

PR #103, opened 2015-08-22
[udev] handle dvb

sledges:
I assume dvb -> dvb (unlike dri -> drm case where name differs)
Devices which will benefit from this fix, do not have dvb under their /sys and /dev even under their Android ROMs, so we can't validate this atm; hence we'll assume it's ok until we add DVB support for Sailfish one day ;)


mlehtima:
LGTM

PR #104, opened 2015-08-22
[mako][pulseaudio] Load BlueZ4 discovery module.

zhxt:
Trying to load module-bluetooth-discover with bluez4 discover module
arguments causes bluez5-discover module loading to fail, which causes
whole module loading to fail. Until module-bluetooth-discover is fixed
to pass only bluez-specific arguments to the discovery modules, just
load module-bluez4-discover directly.


sledges:
LGTM

PR #105, opened 2015-08-25
[c8813q/ g525 ] Add configs for Huawei c8813q / g525-00

drgogeta86:


device-huawei-c8813q-configs/var/lib/environment/compositor/droid-hal-device.conf
```diff
@@ -0,0 +1,5 @@
+# Config for huawei/c8813q
+HYBRIS_EGLPLATFORM=fbdev
```
sledges: please test if fingerterm works. if not, most likely this flag is the reason, and it anyhow needs to be removed (will be fixed in next HADK pdf)

mlehtima:
Device configurations have been for a long time in separate repos so this is obsolete.

PR #110, opened 2015-09-02
[kernel] Add kernel EXTRAVERSION appending for OBS builds.

kjokinie:
In order to get better trackability for kernel, append
+<dhd version> to kernel EXTRAVERSION when we do OBS based builds.

E.g. a previously "3.10.20" versioned kernel will turn into something
like "3.10.20+0.0.6-1".

The code checks also linux/kernel folder in case normal kernel/ folder
does not contain a Makefile. This is to be compatible with Intel android
trees.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -216,6 +216,16 @@ cp %{SOURCE40} %{SOURCE0}
 echo _target_cpu is %{_target_cpu}
 
 %if 0%{?_obs_build_project:1}
+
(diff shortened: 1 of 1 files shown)


kjokinie:
Please note in OBS builds there is no git history, so even if CONFIG_LOCALVERSION_AUTO is enabled, there will not be a git SHA1 appended in version. So the version string will be as specified with in the patch description above.

sledges:
LGTM :+1:

PR #111, opened 2015-09-09
Fixed comment typo

Wnt:
should be t instead of t with stroke


sledges:
Thanks! Was looking at it this morning and cringed :D

PR #113, opened 2015-09-10
[kernel] Add devicetree image to kernel if it exists.

kjokinie:
Device tree is coming with new ARM devices. With this change we package
to kernel package the dt.img binary if it exists.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -435,6 +435,13 @@ KERNEL_RELEASE=$(cat out/target/product/%{device}/*/*/include/config/kernel.rele
 cp out/target/product/%{device}/kernel $RPM_BUILD_ROOT/boot/kernel-$KERNEL_RELEASE
 cp out/target/product/%{device}/obj/ROOT/hybris-boot_intermediates/boot-initramfs.gz $RPM_BUILD_ROOT/boot/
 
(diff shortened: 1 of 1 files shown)


droid-hal-device.inc
```diff
@@ -435,6 +435,13 @@ KERNEL_RELEASE=$(cat out/target/product/%{device}/*/*/include/config/kernel.rele
 cp out/target/product/%{device}/kernel $RPM_BUILD_ROOT/boot/kernel-$KERNEL_RELEASE
 cp out/target/product/%{device}/obj/ROOT/hybris-boot_intermediates/boot-initramfs.gz $RPM_BUILD_ROOT/boot/
 
+echo "/boot/kernel-$KERNEL_RELEASE" > kernel.files
+echo "/boot/boot-initramfs.gz" >> kernel.files
+
+if cp out/target/product/%{device}/dt.img $RPM_BUILD_ROOT/boot/; then
```
saukko: Doesn't this cause errors to be printed if not present? I would either redirect to &> /dev/null or do -f check before cp.

kjokinie: Yes, it does spew out a error message, I'll clean that out.

kjokinie:
Directed the errors from dt cp to /dev/null.

sledges:
LGTM :+1:

PR #114, opened 2015-09-21
[dhd] Fix to prevent always installing -devel with main package.

mlehtima:
Currently -devel package is always installed with main package because of the presence of debug versions of the libraries in -devel. Automatic installation of -devel on devices is not desired so this fix prevents that from happening by excluding the relevant folder from provides.

### DIFF
--- a/helpers/util.sh
+++ b/helpers/util.sh
@@ -135,7 +135,7 @@ function buildmw {
         [ -f "$LOG" ] && rm "$LOG"
 
         if [ ! -d $PKG ] ; then
-            minfo "Source code directory doesn't exist, clonig repository"
+            minfo "Source code directory doesn't exist, cloning repository"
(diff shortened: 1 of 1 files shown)


droid-hal-device.inc
```diff
@@ -37,6 +37,7 @@
 %endif
 
 %define __provides_exclude_from ^%{_libexecdir}/droid-hybris/.*$
+%define __provides_exclude_from ^%{_libdir}/droid-devel/.*$
```
saukko: This overrides the above line, reason for that?

mlehtima: It's a little unclear why the original exclude exists, but after enabling both excludes both the main and -devel packages break. What about replacing the original (which actually happens now) so the libraries would be provided by the main package and not the -devel which is the case currently. This happens at least on local builds of droid-hal-device.

PR #115, opened 2015-09-24
[dhd] Allow device specific requires

sledges:


sledges:
No need, it's as simple as https://github.com/mer-hybris/droid-config-hammerhead/blob/fbc7bce5bf6722ec5090699a9eb908d3d02828fa/rpm/droid-config-hammerhead.spec

PR #118, opened 2015-10-29
Hybris >1.1.9

carepack:
Added cellular components and geoclue-provider-hybris-community to patterns. Disabled the UI inversion -> set value to zero


sledges:
The first commit is already in: https://github.com/mer-hybris/droid-hal-device/commit/078f5f4c7b64025c54693db4701ff99c9db9bd95
The other commits are just history of experiments

Check out clean branch from hybris-10.1, and cherry-pick into it the ones you want, don't forget to prefix them with [mako][<area>]

PR #119, opened 2015-10-29
[mako] modified patterns for cellular functionaliy and disabled the ui inversion

carepack:
since mako inverted the ui the value must be changed to zero. added additional pattern entries for cellular packages and geoclue-provider-hybris-community

Change-Id: I35e463fef8440a16189e256053e782b78574970b


sledges:
I'll accept this only because hybris-10.1 is not maintained any longer, and you promise to split one logical change per one commit and prefix it with "[mako][type]" in the future

PR #121, opened 2015-11-05
Remove kernel modules from droid-hal-device rpm

mkosola:
Remove kernel modules from droid-hal-device if those are installed under %{device}/system/modules. Modules are installed already on kernel modules package.


saukko:
:+1:

philippedeswert:
Makes sense. LGTM

sledges:
LGTM :+1:

PR #123, opened 2015-12-21
[dhd] Exclude provides and requires of all built libraries.

mlehtima:
Current droid-hal-device always pulls in the -devel package during installation and when building dhd for e.g. Fairphone 2 the libraries built by hybris-hal require a lot of other libraries which causes significant amount of building of android libraries that are not built by default during hybris-hal build. Excluding all provides and requires of all built libraries in droid-devel and droid-hybris solves both problems and should not cause problems.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -39,7 +39,8 @@
 %define rpm_vendor %vendor
 %endif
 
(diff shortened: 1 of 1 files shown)


mlehtima:
Not needed anymore.

PR #124, opened 2016-02-14
[build_packages] Add geoclue-providers-hybris.

mlehtima:
Needed for GPS to work in applications.


sledges:
Will we keep this or geoclue-providers-hybris instead?

Also would be great to have pulseaudio and flinger glues too? :-)

mlehtima:
I changed the package to geoclue-providers-hybris. The pulseaudio-modules-droid-glue has problems building locally without patching and because it also needs the packaged audioflingerglue it should not be built by default.

sledges:
LGTM :+1: and thanks for the info

PR #125, opened 2016-03-09
Fix audio header extraction on Android 6.0 based builds.

mlehtima:
Fixes building for Android 6.0 based devices.


sledges:
If no regressions on previous bases, then LGTM

mlehtima:
The check_header_exists prevents regressions by copying the headers from the new location only if it exists.

sledges:
LGTM :+1:

PR #127, opened 2016-03-30
[usergen] Do not remove and re-add already existing users.

ballock:
The logic is:
1. Generate name-id pairs required by droid layer to droid.ids.
2. On installation create all the required user and groups and note their
   names and ids to droid.ids.created.
3. On update cross-check the two files and only add and remove the name-ids
   that have been changed.

ad. 1. usergroupgen.c no longer creates shellscript output, instead it creates
droid.ids which is a simple text file with name-space-id format, one per line.

ad. 2. postinstall script writes down every name-id it managed to create. It
does not write any user creation of which failed.

ad. 3. it is not verified during user/group removal if the user/group was
modified by the user or with a third-party script. If the entry disappeared
from droid-hal, we just remove it.

Special note: A conversion mechanism was added for upgrades from previous dhd
releases in the preinstall script. All users that were added in the previous
release are considered to have been created.

Change-Id: I97cfb9768b6a5edc81590f432d62d7d09539adcb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 helpers/apply-permissions
-helpers/droid-user-add.sh
-helpers/droid-user-remove.sh
+helpers/droid.ids
 helpers/usergroupgen
 
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -284,8 +284,7 @@ ANDROID_ROOT=$(readlink -e %{android_root})
 (cd %{dhd_path}/helpers; make ANDROID_ROOT=$ANDROID_ROOT)
 
 echo Building uid scripts
-%{dhd_path}/helpers/usergroupgen add > %{dhd_path}/helpers/droid-user-add.sh
-%{dhd_path}/helpers/usergroupgen remove > %{dhd_path}/helpers/droid-user-remove.sh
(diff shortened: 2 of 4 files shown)


sledges:
Freshly flashed image with this mod results in stuck at boot Google logo. Files and journal:
https://pastee.org/ddfje
https://pastee.org/4jnk8

ballock:
@sledges I fixed the problem and updated my branch, so it should no longer happen. Also fixed 2 minor bugs, one when the group exists, it used to not create the associated user, the other about ghost file from previous rpm version.

sledges:
This one affects lots of devices and needs testing, please bear with us :)

sledges:
Are they needed with <13.0 ?

ballock:
For a minute I thought I pushed to a wrong repository, but it seems I used my github home's master back in the days to make the pull request.

Apply-permissions change has nothing to do with this particular PR, should be removed from here. But I need my master branch, so will need to move this in some way.

Regarding the original question - no idea. I had them on cm11, no idea if they worked. Jumping to cm13 directly now.

ballock:
Pull request moved to https://github.com/mer-hybris/droid-hal-device/pull/151

PR #136, opened 2016-10-10
 [build_packages] check if $MER_ROOT/devel is existing if not create it and change owner to $USER

Nokius:
This two lines would help new porters not to run into any issue when running the script for the first time.

This step is not covered by the [HADK](https://sailfishos.org/wiki/Hardware_Adaptation_Development_Kit) and can be done in side the script.


Nokius:
obsolete with  HADK 1.9.9

PR #138, opened 2016-10-25
Fix droid-config dependency to droid-config-flashing package

mkosola:
droid-hal-configs commit:
https://github.com/mer-hybris/droid-hal-configs/commit/5359b6791f0e47a617c9eae4177c320d303e7f73


saukko:
LGTM

PR #142, opened 2016-11-04
Hadk polish

sledges:


helpers/util.sh
```diff
@@ -133,7 +133,7 @@ function buildmw {
     GIT_URL="$1"
     shift
     GIT_BRANCH=""
-    if [ "$1" != "" ]; then
+    if [[ "$1" != *.spec ]]; then
```
saukko: I wonder if this should be "*.spec" instead, can shell expand this?

sledges: It doesn't work with "..." :) hence the double [[ ]] https://viewsby.wordpress.com/2013/09/06/bash-string-ends-with/

saukko:
:+1:

PR #144, opened 2016-11-07
Always install a build package on target

sledges:
Otherwise, if rebuilt, it doesn't get pulled in automatically, due to Version: always staying the same on local builds


helpers/util.sh
```diff
@@ -46,6 +47,10 @@ function merror {
 }
 
 function die {
+    if [ "$LOG" != "/dev/null" && -f "$LOG" ] ; then
+        tail -n20 "$1"
+        minfo "Check $1 for full log."
```
saukko: Should these $1 be $LOG?

sledges: They should indeed, thanks! 😄

sledges:
fixed and updated

PR #146, opened 2016-11-18
 [dhd] allow People application to save contacts, let Phone application access contacts, call list

Nokius:
This sed command from @mlehtima fixes the issue on Sonys scorpion, that /tmp has the wrong permissions. Which had effects in the peoples and phone applications.

[source of the sed command ](http://www.merproject.org/logs/%23sailfishos-porters/%23sailfishos-porters.2016-10-01.log.html#t2016-10-01T16:52:18)

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -306,7 +306,7 @@ FSTAB_FILES="$(echo %{android_root}/out/target/product/%{device}/root/fstab.* %{
 shopt -u nullglob
 
 
-%{dhd_path}/helpers/makefstab --files $FSTAB_FILES  --skip auto /acct /boot /cache /data /misc /recovery /staging /storage/usbdisk /sys/fs/cgroup /sys/fs/cgroup/memory /sys/kernel/debug  /sys/kernel/config /dev/usb-ffs/adb %{?makefstab_skip_entries} --outputdir tmp/units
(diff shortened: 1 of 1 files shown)


sledges:
@Nokius , thanks for the PR!

While it's being tested on Jolla devices, please rename the commit message not to say what it does (it's obvious from the diff), but why it does it. And please use the shorter tag [dhd] as per looking at commit history

Nokius:
Sorry for the typo mess

droid-hal-device.inc
```diff
@@ -306,7 +306,7 @@ FSTAB_FILES="$(echo %{android_root}/out/target/product/%{device}/root/fstab.* %{
 shopt -u nullglob
 
 
-%{dhd_path}/helpers/makefstab --files $FSTAB_FILES  --skip auto /acct /boot /cache /data /misc /recovery /staging /storage/usbdisk /sys/fs/cgroup /sys/fs/cgroup/memory /sys/kernel/debug  /sys/kernel/config /dev/usb-ffs/adb %{?makefstab_skip_entries} --outputdir tmp/units
+%{dhd_path}/helpers/makefstab --files $FSTAB_FILES  --skip auto /acct /boot /cache /data /dev/usb-ffs/adb /misc /recovery /staging /storage/usbdisk /sys/fs/cgroup /sys/fs/cgroup/memory /sys/kernel/debug  /sys/kernel/config /tmp %{?makefstab_skip_entries} --outputdir tmp/units
```
sledges: Could you move `/dev/usb-ffs/adb` to where it was originally, so the diff doesn't look confusing?

Nokius: sure I can, the change was done to have all path in alphabetic order. It looked to me it was it once :-s

sledges: Then that would deserve a separate commit (1 logical change per commit :wink:)

PR #148, opened 2016-12-04
[dhd] Skip mountpoint /dev/usb-ffs/adb because it breaks MTP.

mlehtima:
Mounting functionfs to /dev/usb-ffs/adb as done by Android will break MTP in Sailfish OS so let's skip that mountpoint to make MTP work again. This problem appeared after https://github.com/mer-hybris/droid-hal-device/commit/dd9ce1e4ce9ff76726894341e0775a9e86b8e003 fixed the mountpoint name and the mount started to work. At least Fairphone 2 and OnePlus X were affected.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -306,7 +306,7 @@ FSTAB_FILES="$(echo %{android_root}/out/target/product/%{device}/root/fstab.* %{
 shopt -u nullglob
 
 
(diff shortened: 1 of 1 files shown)


kimmoli:
LGTM

tl;dr
It seems that FunctionFS uses mount options of the first mount for the following mounts and ignores options set for them.

In our case `/dev/usb-ffs/adb` options uid=2000,gid=2000 were also used for mount of `/dev/mtp` which causes permission errors.

PR #149, opened 2016-12-13
Build refactorings and community-adaptation enablers

sledges:


mlehtima:
LGTM, tested building packages using all possible methods for Fairphone 2. Also tested installing the needed packages to MerSDK and building per spec.

PR #153, opened 2017-01-24
[dhd] Fix PARANOID_NETWORK automation

sledges:
Previous solution never worked, as it's done in %install, where
%{android_root} no longer exists


mkosola:
:+1:  Tested, works!

sledges:
Thanks @mkosola!

PR #155, opened 2017-02-02
Add mechanism for applying patches to extracted headers.

jusa:


sledges:
Squash the two, so they can be reverted without orphans too? :)

PR #162, opened 2017-05-11
Remove android e2fsprogs tools

neochapay:


saukko:
Why? Your git commmit does not have any reasoning why this is done.

neochapay:
In some android device tree building e2fsprogs hardcored. if we staying put e2fsck and resize2fs we can`t build rootfs becouse droid-hal-device package will be confilct with mer e2fsbrogs

PR #166, opened 2017-08-07
fix audioflingerglue-localbuild

r0kk3rz:
update audioflingerglue-localbuild to match changes, fixing the lib path to allow for 64bit builds


jusa:
Could you change the commit to have tag for changelog, ie
[helpers] Fix audioflingerglue-localbuild

PR #168, opened 2017-08-17
[dhd] Add support for Android 7 based builds.

mlehtima:
WIP: Missing symlinking of init*.rc files.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -315,6 +315,14 @@ hybris/mer-kernel-check/mer_verify_kernel_config \
 # Check if local overrides for tool mk files exist, otherwise dhd defaults are used
 if test -f rpm/helpers/mkbootimg.mk; then
     export MKBOOTIMG_MK=$(pwd)/rpm/helpers/mkbootimg.mk
+else
+    # Starting from Android 7 mkbootimg is a python program so build only for older Android versions
+    if [ $(grep 'PLATFORM_VERSION := ' \
+                          %{android_root}/build/core/version_defaults.mk \
+                          | awk '{print $3}' \
+                          | awk -F'.' '{print $1}') -lt "7" ]; then
+        export MKBOOTIMG_MK=$(pwd)/rpm/dhd/helpers/mkbootimg.mk
+    fi
 fi
 
 if test -f rpm/helpers/simg2img.mk; then
@@ -415,7 +423,7 @@ cp -a %{android_root}/out/target/product/%{device}/hybris-updater-script $RPM_BU
 cp -a %{android_root}/out/target/product/%{device}/hybris-updater-unpack.sh $RPM_BUILD_ROOT/boot
 %endif
 
-cp -a %{android_root}/out/target/product/%{device}/root/. $RPM_BUILD_ROOT/
+rsync -av %{android_root}/out/target/product/%{device}/root/. $RPM_BUILD_ROOT --exclude etc
 
 # We'll use the mer provided modprobe so clean that out of the way
 rm -f $RPM_BUILD_ROOT/sbin/modprobe
@@ -425,6 +433,11 @@ sed -i -e '/^[[:space:]]*mount[[:space:]]/s/^/# Removed during droid-hal-device 
 # remove the creation of /tmp to prevent wrong permissons 
 sed -i -e '/^[[:space:]]*mkdir[[:space:]]\/tmp[[:space:]]*/s/^/# Removed during droid-hal-device build : /' $RPM_BUILD_ROOT/*rc
 
+if [ $ANDROID_VERSION_MAJOR -ge "7" ]; then
+mkdir -p $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system/etc/init
+cp -a %{android_root}/out/target/product/%{device}/system/etc/init/servicemanager.rc $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system/etc/init/
+echo "%{_libexecdir}/droid-hybris/system/etc/init/servicemanager.rc" > tmp/droid-hal.files
+fi
 cp -a %{android_root}/out/target/product/%{device}/system/{bin,lib} $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system/.
 %if 0%{?droid_target_aarch64:1}
 cp -a %{android_root}/out/target/product/%{device}/system/lib64 $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system/.
@@ -467,6 +480,7 @@ if (grep -q '^TARGET_USES_QCOM_BSP := true' %{android_root}/device/%{vendor}/*/B
     cat <<EOF >> $HDRS/android-config.h.new
 /* Added by automatic QCOM_BSP detection */
 #define QCOM_BSP 1
+#define QTI_BSP 1
 EOF
 fi
 sed '0,/CONFIG GOES HERE/d' $HDRS/android-config.h >> $HDRS/android-config.h.new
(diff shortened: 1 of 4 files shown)


droid-hal-device.inc
```diff
@@ -512,7 +517,11 @@ install -D %{dhd_path}/helpers/droid-user-remove.sh $RPM_BUILD_ROOT%{_libdir}/dr
 # Remove cruft
 rm -f $RPM_BUILD_ROOT/fstab.*
 rm -f $RPM_BUILD_ROOT/*goldfish*
+if ([ $ANDROID_VERSION_MAJOR -ge "7" ]); then
+rm -rf $RPM_BUILD_ROOT/{proc,sys,dev,sepolicy} $RPM_BUILD_ROOT/{file,seapp}_contexts
```
sledges: Thanks! Could you move this line out of the if block please?..

droid-hal-device.inc
```diff
@@ -512,7 +517,11 @@ install -D %{dhd_path}/helpers/droid-user-remove.sh $RPM_BUILD_ROOT%{_libdir}/dr
...
+else
 rm -rf $RPM_BUILD_ROOT/{proc,sys,dev,sepolicy} $RPM_BUILD_ROOT/{file,property,seapp}_contexts
```
sledges: ...and here remove only property_contexts?

helpers/build_packages.sh
```diff
@@ -132,6 +132,8 @@ if [ "$BUILDMW_REPO" == "" ]; then
 # hack until upstream is sane
 if (grep -q 'PLATFORM_VERSION := 6.' $ANDROID_ROOT/build/core/version_defaults.mk); then
 buildmw libhybris mm64-rpm || die
+elif (grep -q 'PLATFORM_VERSION := 7.' $ANDROID_ROOT/build/core/version_defaults.mk); then
+buildmw libhybris n64-rpm || die
 else
 buildmw libhybris || die
 fi
```
sledges: down below there is this line:
``buildmw geoclue-providers-hybris jb36857 || die``
I think it's safe to move it back to master? Android 7 certainly needs the latest master

mlehtima: I think the master branch requires Sailfish OS 2.1.1 which is not yet released but maybe we can wait for that release before merging this and fix it here.

sledges: Ah yes I forgot that 😅 The question now remains, are there more porters who are trying out 2.1.1 during EA phase, or those porting towards the released 2.1.0 :)

mlehtima: A good question, probably new porters use 2.1.0 but old porters most likely are moving to 2.1.1. But old porters often use OBS so this won't affect that.

sledges: Oki, let's leave the _status quo_ as is then :)

droid-hal-device.inc
```diff
@@ -425,6 +425,10 @@ sed -i -e '/^[[:space:]]*mount[[:space:]]/s/^/# Removed during droid-hal-device
 # remove the creation of /tmp to prevent wrong permissons 
 sed -i -e '/^[[:space:]]*mkdir[[:space:]]\/tmp[[:space:]]*/s/^/# Removed during droid-hal-device build : /' $RPM_BUILD_ROOT/*rc
 
+if ([ $ANDROID_VERSION_MAJOR -ge "7" ]); then
```
sledges: nitpick: (brackets) are not needed

droid-hal-device.inc
```diff
@@ -512,7 +517,10 @@ install -D %{dhd_path}/helpers/droid-user-remove.sh $RPM_BUILD_ROOT%{_libdir}/dr
 # Remove cruft
 rm -f $RPM_BUILD_ROOT/fstab.*
 rm -f $RPM_BUILD_ROOT/*goldfish*
-rm -rf $RPM_BUILD_ROOT/{proc,sys,dev,sepolicy} $RPM_BUILD_ROOT/{file,property,seapp}_contexts
+rm -rf $RPM_BUILD_ROOT/{proc,sys,dev,sepolicy} $RPM_BUILD_ROOT/{file,seapp}_contexts
+if ([ $ANDROID_VERSION_MAJOR -lt "7" ]); then
```
sledges: same as above

droid-hal-device.inc
```diff
@@ -415,7 +415,7 @@ cp -a %{android_root}/out/target/product/%{device}/hybris-updater-script $RPM_BU
 cp -a %{android_root}/out/target/product/%{device}/hybris-updater-unpack.sh $RPM_BUILD_ROOT/boot
 %endif
 
-cp -a %{android_root}/out/target/product/%{device}/root/. $RPM_BUILD_ROOT/
+rsync -av %{android_root}/out/target/product/%{device}/root/. $RPM_BUILD_ROOT --exclude etc
```
sledges: what if we don't want all .rc files over? like no netd/bootanimation/...?

droid-hal-device.inc
```diff
@@ -425,6 +425,10 @@ sed -i -e '/^[[:space:]]*mount[[:space:]]/s/^/# Removed during droid-hal-device
 # remove the creation of /tmp to prevent wrong permissons 
 sed -i -e '/^[[:space:]]*mkdir[[:space:]]\/tmp[[:space:]]*/s/^/# Removed during droid-hal-device build : /' $RPM_BUILD_ROOT/*rc
 
+if [ $ANDROID_VERSION_MAJOR -ge "7" ]; then
+mkdir -p $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system/etc
+cp -a %{android_root}/out/target/product/%{device}/system/etc/init $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system/etc/.
```
sledges: ^ the above comment should've been here :)

mlehtima: At least on my hybris-14.1 build only servicemanager.rc appears in that folder but of course we could change it so that only that file is copied because we haven't modified any other .rc files that might go there.

sledges: Ah yes, here only "rebuilt" .rc files will appear and get copied in. But haven't we forgot the fallback to /system/etc/init then https://github.com/mlehtima/android_system_core/commit/5b9ccc722464eddf92e807e51dd8d5362e5a054a#diff-aedeb8c2e4984c93564cb58ce327f103L528 ?

mlehtima: There is no fallback to /system/etc/init, we replaced that with /usr/libexec/droid-hybris/system/etc/init so we need to have all services we need in that folder, hopefully we get the symlinking working so we can just copy the servicemanager.rc and symlink all other needed services.

sledges: That's what I mean, all other needed .rc files are not in the out/ after make hybris-hal. Having them via droid-configs defeats the purpose of this copy in dhd (and also might conflict of same file in two packages)

mlehtima: I thought the symlinking was supposed to be done in this script, not in droid-configs. But it is true that only servicemanager.rc should be copied to rule out any conflicts when symlinking.

sledges: Yes we need to be careful with that script, as /system might not be yet mounted during install of dhd (especially in mic phase). Are you having some systemd .service in mind that calls a shell script on every boot that checks the state of things need to be copied/symlinked?

mlehtima: I meant to do the symlinking when building droid-hal i.e. in this build script, assuming of course we get the symlinking working. This way the symlinks to the needed .rc files are always added and present when droid-hal-init is run. However is the symlinking cannot be fixed then a systemd service might be needed.

sledges: Ok that's cool too; yet different vendors/versions might have different sets, we could of course add the full set and keep on updating it (just like with /etc/*.xml symlinks).

Could we merge this PR and relax on implementing/testing the symlinks script, for now the adaptations could just add symlinks/files to droid-configs sparse? and later on remove it when this is up?

mlehtima: We can do that, I'll just modify the script to only copy servicemanager.rc for now and we can then later decide how to add the symlinking/copying of .rc files. As for vendor .rc files, those should be in /vendor/etc/init which is also used by init.

mlehtima:
There is still the issue of img2simg and simg2img building which was excluded from this PR: Those need to be disabled in Android 7 builds at the moment

mlehtima:
Building of img2simg and simg2ing is now fixed, the problem was change of mkbootimg from a c program to a python program which caused the build of it to fail and therefore also img2simg and simg2img which were built after that.

droid-hal-device.inc
```diff
@@ -315,6 +315,13 @@ hybris/mer-kernel-check/mer_verify_kernel_config \
 # Check if local overrides for tool mk files exist, otherwise dhd defaults are used
 if test -f rpm/helpers/mkbootimg.mk; then
     export MKBOOTIMG_MK=$(pwd)/rpm/helpers/mkbootimg.mk
+else
+    if [ $(grep 'PLATFORM_VERSION := ' \
```
sledges: A comment would be handy here explaining that starting with Android 7 mkbootimg is a Python script, and no longer requires make/building (IIUC)

sledges:
Thanks for the last chg req., sorry for late merge, I wish github sent email alerts when a push to PR is made, just like bitbucket does :)

PR #173, opened 2017-09-19
[dhd] Ignore some unused mount points.

mlehtima:


droid-hal-device.inc
```diff
@@ -363,7 +363,7 @@ FSTAB_FILES="$(echo %{android_root}/out/target/product/%{device}/root/fstab.* %{
 shopt -u nullglob
 
 
-%{dhd_path}/helpers/makefstab --files $FSTAB_FILES  --skip auto /acct /boot /cache /data /misc /recovery /staging /storage/usbdisk /sys/fs/cgroup /sys/fs/cgroup/memory /sys/kernel/debug  /sys/kernel/config /dev/usb-ffs/adb /tmp %{?makefstab_skip_entries} --outputdir tmp/units
+%{dhd_path}/helpers/makefstab --files $FSTAB_FILES  --skip auto none /acct /boot /cache /data /misc /recovery /staging /storage/sdcard1 /storage/usbdisk /storage/usbotg /sys/fs/cgroup /sys/fs/cgroup/memory /sys/kernel/debug  /sys/kernel/config /dev/usb-ffs/adb /tmp %{?makefstab_skip_entries} --outputdir tmp/units
```
sledges: sdcard bits need to be tested on official devices that have SD card (testing whether there is a storage-sdcard.mount file is sufficient I think, no need to test with actual SD card)

mlehtima: No storage-sdcard.mount files found on Jolla 1, Jolla C, Jolla Tablet or Xperia X. Any other devices need to be checked by someone who has them. Also sdcard is mounted based on udev rules included in sd-utils https://git.merproject.org/mer-core/sd-utils/blob/master/rules/90-mount-sd.rules

sledges: Break things to make things then :)

PR #175, opened 2017-09-26
OBS build fixes

sledges:


sledges:
One way to send email alerts about updated PR 😅 

Synced OBS macro variable name in-line with https://github.com/mer-hybris/droid-hal-source/pull/1

PR #177, opened 2017-11-09
Add support of outdour fixup-mountpoints

neochapay:
Needed for https://github.com/mer-hybris/droid-hal-device/pull/164


sledges:
This is not needed; $ANDROID_ROOT/hybris/hybris-boot/fixup-mountpoints will always exist, regardless if hybris-boot is used as submodule or not.

If used as submodule:
```
hybris
├── hybris-boot/
|   ├── fixup-mountpoints (must exist and is used, otherwise don't use submodule)
|   ├── hb/ (submodule)
|       ├── fixup-mountpoints (not used)
|       ├── ... (other files)
```
If not used as submodule
```
hybris
├── hybris-boot/
|   ├── fixup-mountpoints (used)
|   ├── ... (other files)
```

neochapay:
From rpm build we ALWAYS use hybris/hybris-boot/fixup-mountpoints when build rpm

sledges:
exactly, and that's how it is going to stay

PR #178, opened 2017-11-16
For new libhybris

sledges:
Instructions will be relevant when https://github.com/mer-hybris/libhybris/pull/35 is merged


krnlyng:
lgtm

PR #183, opened 2018-04-16
Allow specific android-boot modes to be preserved. JB#41430

krnlyng:
This is in order to allow booting into vendor specific modes, which should run as close as possible to the original Android version.


droid-hal-device.inc
```diff
@@ -457,10 +459,15 @@ rsync -av %{android_root}/out/target/product/%{device}/root/. $RPM_BUILD_ROOT --
 # We'll use the mer provided modprobe so clean that out of the way
 rm -f $RPM_BUILD_ROOT/sbin/modprobe
 
+%if %{?initrc_skip_files:1}%{!?initrc_skip_files:0}
```
sledges: Simpler:
```%if 0%{?initrc_skip_files:1}```

krnlyng:
Not used anymore, we can close if nobody else requires a similar thing.

sledges:
Closing, thanks for experimenting though!:)

PR #185, opened 2018-06-07
[dhd] Fix build for Android 8.

krnlyng:
Adapt for new paths and changes in android 8.

Co-authored-by: Franz-Josef Haider <franz.haider@jolla.com>

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -377,6 +377,8 @@ else
     echo "#include <private/android_filesystem_config.h>;" > generated_android_filesystem_config.h
 fi
 
+USE_STD_C99=1
+
 # files have changed in android >=8 so we need to supply different ones to the
 # makefile. we also need the cstring workaround thus handling both issues here
 # at once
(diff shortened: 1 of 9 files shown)


helpers/makefstab
```diff
@@ -25,6 +25,45 @@ while (@ARGV) {
...
+
+# From $fil :
+# $cont
+END
+  print "Wrote $uname.unit\n";
+  close UNIT;
+}
+
 for my $file (@files) {
   open(FILE, "<$file") or die "Can't read from $file\n";
```
mlehtima: If building for older android or device bases which do not have /vendor/etc/fstab* files the script should not die here, it should instead have "or next;" to skip not found files.

krnlyng: @mlehtima the FSTAB_FILES glob had a wrong bracket, this should now be fixed.

mlehtima:
LGTM, atfer the last small fix droid-hal built without issues in both Android 6 and 8 bases.

droid-hal-device.inc
```diff
@@ -364,7 +368,33 @@ fi
...
+    echo "#include \"sparse_read.cpp\"" >> $ANDROID_ROOT/system/core/libsparse/sparse_read_fix.cpp
+    IMG2SIMG_SOURCES="backed_block.c output_file.c sparse.c sparse_crc32.c sparse_err.c sparse_read_fix.cpp img2simg.c ../base/stringprintf.cpp"
+    SIMG2IMG_SOURCES="backed_block.c output_file.c sparse.c sparse_crc32.c sparse_err.c sparse_read_fix.cpp simg2img.c ../base/stringprintf.cpp"
+else
+    IMG2SIMG_SOURCES="backed_block.c output_file.c sparse.c sparse_crc32.c sparse_err.c sparse_read.c img2simg.c"
+    SIMG2IMG_SOURCES="backed_block.c output_file.c sparse.c sparse_crc32.c sparse_err.c sparse_read.c simg2img.c"
+fi
+
+pushd %{dhd_path}/helpers
+make ANDROID_ROOT=$ANDROID_ROOT IMG2SIMG_SOURCES="$IMG2SIMG_SOURCES" SIMG2IMG_SOURCES="$SIMG2IMG_SOURCES"
```
sledges: Minor: some adaptations override the .mk files, but mkbootimg is the only I could quick-find https://github.com/mer-hybris/droid-hal-mako/blob/master/helpers/mkbootimg.mk (yet still good to bear in mind we might have a porter or two knocking at door due to this API change :) )

krnlyng: @sledges as far as i can tell overwriting *.mk files should still work :). If issues arise we need to tackle those later yes :).

krnlyng:
Added another commit on top, to allow middleware packages to be precloned to external. This is a requirement for the surfaceflinger approach, otherwise either local builds will fail, or obs builds will fail due to awkward dependency of hybris-hal on libhybris and dhd on hybris-hal :) and libhybris on dhd. If we can change libhybris to have a seperate libhybris-compat repository instead of having a compat folder this would help but we need to confirm with upstream if this is ok.

PR #187, opened 2018-07-12
Set USE_STD_C99=0 for libsparse in order to avoid build failure

RedL0tus:
Otherwise it will throw an error while compiling, not just for Android >= 8

Build log for a failed build: http://termbin.com/6i3r

Huge thanks to @zhxt

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -390,11 +390,12 @@ if [ $ANDROID_VERSION_MAJOR -ge "8" ]; then
     IMG2SIMG_SOURCES="backed_block.c output_file.c sparse.c sparse_crc32.c sparse_err.c sparse_read_fix.cpp img2simg.c ../base/stringprintf.cpp"
     SIMG2IMG_SOURCES="backed_block.c output_file.c sparse.c sparse_crc32.c sparse_err.c sparse_read_fix.cpp simg2img.c ../base/stringprintf.cpp"
 
-    USE_STD_C99=0
 else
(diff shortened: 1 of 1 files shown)


krnlyng:
Is this required for all adaptations? Android 4 - 7/8?

RedL0tus:
@krnlyng 
Well... I'm not sure, but at leaset it's required for Android 7.
The problem is libsparse uses a GNU99 style `typeof` in its `min` macro, and it will cause it fail to compile if the compiler use standard C99.
https://github.com/mer-hybris/android_system_core/blob/hybris-10.1/libsparse/output_file.c#L59
And the `min` macro is the same from CM10.1 to 15.1.

The solution is to set `USE_STD_C99=0`  or rewrite that macro. I saw someone said he rewrote that macro but he didn't pushed it to anywhere, so...

krnlyng:
@RedL0tus well currently the USE_STD_C99 was used for android < 8. But the patch you made still has the same detection for USE_STD_C99 and it then unconditionally overwrites it which is a bit weird.
I think the problem is that we had to switch the compiler to g++ for android 8.
In my opinion is to either, get rid of the whole USE_STD_C99 code, or if that is not possible (due to android code?) then when USE_STD_C99 is set, also use gcc instead of g++.

RedL0tus:
@krnlyng Oh... I'm afraid that I cannot help on that since I'm not familiar with C/C++. I will close this PR and open an issue instead (since it is still a problem that will cause build failure).

PR #191, opened 2018-08-20
Unbreak zypp and other niggles

sledges:


mlehtima:
LGTM, tested build using 2.2.0.29 target.

PR #192, opened 2018-09-26
[build] Fix geoclue-providers-hybris build. JB#42318

krnlyng:


krnlyng:
hm. this won't work on the first build (only on successive builds...)

krnlyng:
should be better now.

helpers/build_packages.sh
```diff
@@ -136,6 +136,9 @@ pushd $ANDROID_ROOT/hybris/mw > /dev/null
 
 if [ "$BUILDMW_REPO" == "" ]; then
 buildmw libhybris || die
+# version can be detected here, earlier is not possible since libhybris is the
+# first thing which BuildRequires: pkgconfig(android-headers)
```
sledges: I haven't looked properly the first time around, and thought that this was after dhd-devel package is pulled in (https://github.com/mer-hybris/droid-hal-device/blob/74c6f315c8bd6d529f04d2b84944daf3031e2a0b/helpers/build_packages.sh#L133), which is also ok anywhere after that line.
LGTM here too

krnlyng: brain malfunction :)

PR #193, opened 2018-10-25
SELinux: Do not remove SELinux-related files

ibaranovomp:


ibaranovomp:
That was not my decision to remove seapp_contexts and i have no any motivation for revisit it.
This file not used now.

PR #196, opened 2018-11-28
Add support of custom fixup-mount points. Fixes jb43993

Shiwin:


sledges:
This can be achieved by using hybris-boot as submodule: https://github.com/mer-hybris/hybris-boot/commit/bac86153d88746825c54d00bba6f212405b01d50

PR #205, opened 2019-02-26
Rework buildmw() to use named arguments

ix5:
## build_packages.sh
Allow supplying multiple `.spec` files

## `buildmw()`: Use named parameters

Instead of relying on positional arguments, allow the following:
- `-u` for git URL
- `-b` for branch name
- `-s` for spec file. Can be supplied multiple times

## mw: Use named opts to call buildmw()

Using named arguments for calling `buildmw()` makes the whole thing more readable.
Use full URLs as well for better readability.

## `build()`: Consider all supplied spec files

Using `$@` instead of `$1` means all arguments are iterated over when looping `$SPECS`

### DIFF
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -42,6 +42,7 @@ function usage() {
     echo "  -v, --version   build droid-hal-version"
     echo "  -b, --build=PKG build one package (PKG can include path)"
     echo "  -s, --spec=SPEC optionally used with -m or -b"
+    echo "                  can be supplied multiple times to build multiple .spec files at once"
     echo " No options assumes building for all areas."
     exit 1
 }
@@ -62,6 +63,7 @@ if [ "$#" == "1" ]; then
     BUILDVERSION=1
 fi
 
+BUILDSPEC_FILE=()
 while true; do
     case "$1" in
       -h|--help) usage ;;
@@ -79,7 +81,7 @@ while true; do
           shift;;
       -s|--spec) BUILDSPEC=1
           case "$2" in
-              *) BUILDSPEC_FILE=$2;;
+              *) BUILDSPEC_FILE+=("$2");;
           esac
           shift;;
       -v|--version) BUILDVERSION=1 ;;
@@ -110,7 +112,8 @@ if [ "$BUILDCONFIGS" == "1" ]; then
         ret=$?
         if [ $ret -eq 104 ]; then
             BUILDALL=y
-            buildmw https://github.com/mer-hybris/community-adaptation.git rpm/community-adaptation-localbuild.spec || die
+            buildmw -u "https://github.com/mer-hybris/community-adaptation.git" \
+                    -s rpm/community-adaptation-localbuild.spec || die
             BUILDALL=n
         elif [ $ret -ne 0 ]; then
             die "Could not determine if community-adaptation package is available, exiting."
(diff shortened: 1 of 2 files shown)


helpers/build_packages.sh
```diff
@@ -79,7 +80,7 @@ while true; do
           shift;;
       -s|--spec) BUILDSPEC=1
           case "$2" in
-              *) BUILDSPEC_FILE=$2;;
+              *) BUILDSPEC_FILE+=("$2");;
```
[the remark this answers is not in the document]
sledges: Thank you for your contribution. In the past we never had a case of building multiple .spec files. However now with the Sailfish X Android 8 adaptations we will need a mechanism to do so, and that is exactly what you have just implemented, cheers!

The historical reason why all possible .spec files used to be built by default (https://github.com/mer-hybris/droid-hal-device/blob/8bb1f91a7f0bee0fd165d304f36ffd7e1d5b7a0c/helpers/util.sh#L230) remains hidden with the community person who initially helped writing these build scripts.

helpers/build_packages.sh
```diff
@@ -142,38 +144,51 @@ if [ "$BUILDMW" == "1" ]; then
...
             # pull device's bluez4 configs correctly
             sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -m sdk-install -R zypper remove bluez-configs-mer
         fi
     else
-        buildmw $BUILDMW_REPO $BUILDSPEC_FILE || die
+        if [[ -z "$BUILDSPEC_FILE" ]]; then
+            buildmw -u $BUILDMW_REPO || die
+        else
+            # Supply all given spec files from $BUILDSPEC_FILE array prefixed with "-s"
+            buildmw -u $BUILDMW_REPO "${BUILDSPEC_FILE[@]/#/-s }" || die
```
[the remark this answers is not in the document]
sledges: :+1:

helpers/fix_dangling_adaptation-community_repo.sh
```diff
@@ -16,12 +16,12 @@ if [ -n "$(grep '%define community_adaptation' $ANDROID_ROOT/hybris/droid-config
         echo "adaptation-community has been found pointing to an OBS repo, let's purge it for local builds"
         sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -m sdk-install -R rpm -e --nodeps community-adaptation-devel
         BUILDALL=y
-        buildmw https://github.com/mer-hybris/community-adaptation.git rpm/community-adaptation-localbuild.spec || die
+        buildmw -u "https://github.com/mer-hybris/community-adaptation.git" -s rpm/community-adaptation-localbuild.spec || die
```
sledges: This file is an old relic, and will be removed altogether in a separate commit.

helpers/build_packages.sh
```diff
@@ -187,7 +202,11 @@ if [ "$BUILDPKG" == "1" ]; then
     if [ -z $BUILDPKG_PATH ]; then
        echo "--build requires an argument (path to package)"
     else
-        buildpkg $BUILDPKG_PATH $BUILDSPEC_FILE
+        if [[ -z "$BUILDSPEC_FILE" ]]; then
```
sledges: I don't see the need to check if it's empty...

helpers/build_packages.sh
```diff
@@ -187,7 +202,11 @@ if [ "$BUILDPKG" == "1" ]; then
...
+            buildpkg $BUILDPKG_PATH
+        else
+            buildpkg $BUILDPKG_PATH "$BUILDSPEC_FILE"
```
sledges: ...as this should suffice (previously it would pass only the first argument):
`buildpkg $BUILDPKG_PATH ${BUILDSPEC_FILE[@]}`

This https://github.com/mer-hybris/droid-hal-device/pull/205/files?utf8=%E2%9C%93&diff=unified#diff-748fa8f198703f775c5df5dc1fa219ccR295 then should be corrected as:
```
shift
build $@
```

ix5: Good catch, I've incorporated your suggestion now.

helpers/build_packages.sh
ix5: Dropped

helpers/fix_dangling_adaptation-community_repo.sh
ix5: Well I don't want to change any real behaviour in this commit. If you push any change I will rebase and incorporate it, but I would like to keep this PR clean.

sledges: Just pushed to master, let's keep this PR also minimal :)

ix5: Rebased and deleted.

ix5:
@sledges If you could do a test run on your machine as well to make sure I haven't overlooked any grave errors, I think this should be ready for merge.

sledges:
I've already tested it; please squash all commits in one so it reflects one logical change, avoiding breakage if accidentally only one of the commits gets reverted.

ix5:
Squashed and force-pushed

sledges:
Many thanks. Since you didn't know that we never built multiple .spec files, your commit message describes of what has been done to fix/implement this functionality.

Could you please instead word it in such a way so that readers of the commit list would instantly know that this functionality has now been introduced, e.g. allow building multiple .spec files.

A documentation update would also be appreciated to (e.g. add "can be repeated to build multiple .spec files"):
https://github.com/mer-hybris/droid-hal-device/blob/8bb1f91a7f0bee0fd165d304f36ffd7e1d5b7a0c/helpers/build_packages.sh#L44

I have added the following commit to fix the build of Xperia XA2:
https://github.com/sledges/droid-hal-device/commit/c8b9c559e9a67ddfab9e8a8e53e0d6508caeca69

Full instructions how to build XA2 (including utilising multiple .spec builds) will be made available after this PR is merged (similar to https://sailfishos.org/wiki/Sailfish_X_Build_and_Flash which was only for the Xperia X series)

ix5:
> Many thanks. Since you didn't know that we never built multiple .spec files, your commit message describes of what has been done to fix/implement this functionality.
> Could you please instead word it in such a way so that readers of the commit list would instantly know that this functionality has now been introduced, e.g. allow building multiple .spec files.

Done, new commit msg is `Rework buildmw to allow multiple spec files`

> A documentation update would also be appreciated to (e.g. add "can be repeated to build multiple .spec files"):

Done. Help now reads:
>  -s, --spec=SPEC optionally used with -m or -b"
>                              can be supplied multiple times to build multiple .spec files at once"

---

> I have added the following commit to fix the build of Xperia XA2:
> [sledges@c8b9c55](https://github.com/sledges/droid-hal-device/commit/c8b9c559e9a67ddfab9e8a8e53e0d6508caeca69)
> 
> Full instructions how to build XA2 (including utilising multiple .spec builds) will be made available after this PR is merged (similar to https://sailfishos.org/wiki/Sailfish_X_Build_and_Flash which was only for the Xperia X series)

That's great to hear.

ix5:
Thanks for reviewing and merging!

PR #206, opened 2019-02-27
Add build_bootimg helper script

ix5:
This helps with `droid-hal-img-boot` packages that need `yamui`, `e2fsprogs` and more, c.f. [hybris-initrd/droid-hal-device-img-boot.inc](https://github.com/mer-hybris/hybris-initrd/blob/ae58855e8bf911bbf6836e351da1677ec6f49129/droid-hal-device-img-boot.inc#L67-L74)

Didn't want to clutter `build_packages.sh` any further so it's split out into another script.

Dependent on #205 because it uses the named-arg `buildmw -u $URL -s $SPEC` syntax.

### DIFF
--- a/helpers/build_bootimg_packages.sh
+++ b/helpers/build_bootimg_packages.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+# build_bootimg_packages.sh - Build and/or install extra packages
+# needed for droid-hal-img-boot
+
+if [[ ! -d rpm/dhd ]]; then
+    echo $0: 'launch this script from the $ANDROID_ROOT directory'
+    exit 1
+fi
+# utilities
+. ./rpm/dhd/helpers/util.sh
+
+BUILDALL=y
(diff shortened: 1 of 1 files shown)


helpers/build_bootimg_packages.sh
```diff
@@ -0,0 +1,21 @@
...
+    echo $0: 'launch this script from the $ANDROID_ROOT directory'
+    exit 1
+fi
+# utilities
+. ./rpm/dhd/helpers/util.sh
+
+sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -R -msdk-install zypper ref -f
+
+sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -R -m sdk-install zypper -n install -f \
+    e2fsprogs openssh-clients openssh-server
```
sledges: Thank you for your contribution.

There is no need to refresh and install, as these packages are already available and will be pulled in automatically by the builds below. I'll push the fix to your branch and merge this PR.

PR #208, opened 2019-02-28
[dhd] Relax requirement on droid-config-flashing

ix5:
Trust `droid-config-%{rpm_device}-flashing` to provide `droid-config-flashing`.
Eases handling of product families like Sony's lineup.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -228,7 +228,7 @@ Requires: oneshot
 Summary: Boot img for droid-hal device: %{rpm_device}
 %if 0%{?enable_kernel_update:1}
 # The device flashing info and scripts is now in droid-configs
-Requires(post): droid-config-%{rpm_device}-flashing
+Requires(post): droid-config-flashing
 %endif
 
 %description img-boot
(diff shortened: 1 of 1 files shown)


sledges:
Sadly this would break our OBS builds where we build for more than one Sony model and variant (currently we build for 6 of them) and OBS would not know which droid-config-flashing to choose, as there would be 6 to choose from. This could we worked around via OBS's project-config (prjconf), but it'd then be one more place to remember when we add a new device.
What is your use case, and is there a possible compromise?

ix5:
d-h-d wants `droid-config-%{rpm_device}-flashing`, where `rpm_device` will be the model *codename*(`discovery`/`kagura`), not model *number*(`h4331`/`f8331`).
So d-h-d wants `droid-config-kagura-flashing`.

d-h-c provides `droid-config-%{rpm_device}-flashing`, where `rpm_device` is the model *number* this time, e.g. `droid-config-f8331-flashing`. So there's a mismatch.

Because you are setting `have_custom_img_boot` and thus skipping over this section and using `droid-hal-img-boot-*`(which only wants the generic `droid-config-flashing`) you don't see the issue I guess.

ix5:
I realize I'm an edge-case building a product-family device without `droid-hal-img-boot-$DEVICE`.

ix5:
It's all a bit weird, because in other places, you have this:
`rpm_device`: `f8331`
`device`: `kagura`

But in d-h-d you have:
`rpm_device`: `kagura` (because `rpm_device` = `device` if `rpm_device` is empty)
`device`: `kagura`

ix5:
Pfff

PR #209, opened 2019-02-28
[WIP] makefstab: Handle legacy /system/vendor mount

ix5:
Legacy devices might not have a dedicated `/vendor` partition and as such need to bind-mount `/system/vendor` to `/vendor`.


ix5:
Seems this is already accomplished through the symlink in sailfish's rootfs.

PR #211, opened 2019-03-01
[helpers] Allow set $pkg ver in audiofingerglue tar. Fixes MER#2018

Thaodan:
Support setting the version of the audiofingerglue tar archive in dhd
like droidmedia.

### DIFF
--- a/helpers/audioflingerglue-localbuild.spec
+++ b/helpers/audioflingerglue-localbuild.spec
@@ -9,7 +9,7 @@
 
 Name:          audioflingerglue
 Summary:       Android AudioFlinger glue library
-Version:       0.0.1
+Version:       0.0.0
 Release:       1
 Group:         System/Libraries
 License:       ASL 2.0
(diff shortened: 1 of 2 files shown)


helpers/audioflingerglue-localbuild.spec
```diff
@@ -9,7 +9,7 @@
 
 Name:          audioflingerglue
 Summary:       Android AudioFlinger glue library
-Version:       0.0.1
+Version:       0.0.0
```
sledges: This would require a new release of the HADK document, unless you left this as 0.0.1

Thaodan: Just wanted to be consistent. Shouldn't this enforce the requirement to get the git tag of the commit to get the pkg version?

sledges: HADK 3.0.1 is just around the corner, so it will all be consistent now :)
I'm rewording your commit message a little, will push to your branch and merge. Thanks!

sledges: Hm it doesn't allow me to push to your branch, even at the bottom of this PR is written "Add more commits by pushing to the audiofinger_localbuild_version branch on Thaodan/droid-hal-device."
Most probably it's because you forked not immediately from upstream. I'll merge it as-is :)

PR #212, opened 2019-03-02
[helpers] Add option to explicitly don't install a pkg

Thaodan:
Add an option to explicitly don't instal a pkg while deploying it.
This allows to build and deploy pkgs where not every sub-pkg is to
be installed. For example like updating ofono where there's a default
config, that conflicts with the device config or droid-vendor where
setcap fails.

### DIFF
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -47,7 +47,7 @@ function usage() {
     exit 1
 }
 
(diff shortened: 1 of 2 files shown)


ix5:
Is it too late to ask that `--dont-install`be changed to `do-not-install` ?
Since internally `do_not_install`/`DO_NOT_INSTALL` is used, consistency would be nice.

sledges:
It's never too late :) https://github.com/mer-hybris/droid-hal-device/commit/456a3f55c5d7e4b030c9cd89114147972e4cff67

PR #214, opened 2019-03-13
[dhd] Allow skip mer_kernel_check

neochapay:


sledges:
Thanks for your contribution. However wouldn't it be better to update mer_kernel_check by relaxing some config flags that are failing for you?
In this way, you would do a favour to anyone who will have the same errors in the future - the dhd build will just work for them :)

neochapay:
@sledges for some experements like selinux build and etc i must disable mer_kernel_check.

PR #219, opened 2019-06-14
[dhd] Helpers cleanup

Thaodan:
- Use builtin $PWD
- Fix quoting in case path names contain spaces [SC2046]
- Remove useless function keyword 
- Remove useless true just return zero
- Use if directly, using test is redundant
- Remove yesnoal, add --build-all
- Quote ${BUILDSPEC_FILE[@]} to avoid resplitting of elements
- Don't use bash == on posix test [] and use [] everywhere

### DIFF
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -214,7 +214,7 @@ if [ "$BUILDPKG" == "1" ]; then
     if [ -z $BUILDPKG_PATH ]; then
        echo "--build requires an argument (path to package)"
     else
-        buildpkg $BUILDPKG_PATH ${BUILDSPEC_FILE[@]}
+        buildpkg $BUILDPKG_PATH "${BUILDSPEC_FILE[@]}"
     fi
 fi
 
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -33,7 +33,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 
-function usage() {
+usage() {
     echo "Usage: $0 [OPTION]..."
     echo "  -h, --help      you're reading it"
     echo "  -d, --droid-hal build droid-hal-device (rpm/)"
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -102,17 +102,17 @@ if [ $# -ne 0 ]; then
     exit 1
 fi
 
-if [[ ! -d rpm/dhd ]]; then
+if [ ! -d rpm/dhd ]; then
     echo $0: 'launch this script from the $ANDROID_ROOT directory'
     exit 1
 fi
 # utilities
 . ./rpm/dhd/helpers/util.sh
 
-if [ "$BUILDDHD" == "1" ]; then
+if [ "$BUILDDHD" = "1" ]; then
     builddhd
 fi
-if [ "$BUILDCONFIGS" == "1" ]; then
+if [ "$BUILDCONFIGS" = "1" ]; then
     if [ -n "$(grep '%define community_adaptation' $ANDROID_ROOT/hybris/droid-configs/rpm/droid-config-$DEVICE.spec)" ]; then
         sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -m sdk-install -R zypper se -i community-adaptation > /dev/null
         ret=$?
@@ -134,7 +134,7 @@ if [ "$BUILDCONFIGS" == "1" ]; then
     buildconfigs
 fi
 
-if [ "$BUILDMW" == "1" ]; then
+if [ "$BUILDMW" = "1" ]; then
     sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -R -msdk-install ssu domain sales
     sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -R -msdk-install ssu dr sdk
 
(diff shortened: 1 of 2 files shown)


Thaodan:
rebased against master

helpers/build_packages.sh
```diff
@@ -83,6 +83,9 @@ while true; do
               *) BUILDPKG_PATH=$2;;
           esac
           shift;;
+      --build-all)
+          BUILDALL=1
```
sledges: When `--build-all` is not specified in command line, the script still builds all packages.

Also please mention this new flag in the `usage()`.

Thaodan:
Rebased and changed as requested.

helpers/build_packages.sh
sledges: Thanks for rebase, I took those changes and ran the script without --build-all again, and it still builds all packages:)

mlehtima:
Nothing uses the BUILDALL variable in the code after you removed yesnoall, I doubt that is intended? Also it might not be a good idea to completely remove the possibility to skip building certain packages as often porters have to skip packages because they have to build manually (build_packages.sh -b) an older version of some packages due to incompatibility with current Sailfish OS version. Maybe changing yesnoall to yesno or just keeping it would make more sense.

Thaodan:
It was because I've missed add another fix after that.
I refaktored build all  to just build every package (mw, hal and so on).
My original ideas to remove the questions every time a package is build 
with -b and keep the old behaivior when --build-all is used.

I for my self prefer to use --build every time to build pkgs in the order I want 
and to have a better overview in my build system as shown here:
https://gitlab.com/Thaodan/hadk_tools/blob/kumano/templates/
mer.droidmedia.hadk[1] 

--------
[1] https://gitlab.com/Thaodan/hadk_tools/blob/kumano/templates/
mer.droidmedia.hadk

sledges:
This has become too convoluted. BTW your last version of the yesnoall-removing commit has reverted all previous PR's commits.

And now that I understand what you originally meant, the "--build-all" flag, it should've be called "--build-all-mw", because "building of all" packages (dhd+cfg+mw+ver) is already there by default when not providing any cmdline arguments.

Please remove the "yesnoall" commit from this PR (because it is called 'Helpers cleanup') so the cleanup commits could be merged, and please create a separate PR for such functionality change.

The more useful behaviour to reduce the usage of yesnoall IMHO could be:
- `build_packages.sh` (no arguments) = builds all packages and doesn't ask yesnoall during mw (=unassisted build)
- `build_packages.sh --mw` = builds mw packages and asks for yesnoall
- `build_packages.sh --build` = doesn't ask the superfluous singleton yesnoall question
This means no changes to the actual `usage()`.

Comments to this idea welcome.

Thaodan:
I like the idea but why ask during --mw builds?

sledges:
Found this in the commit body 3a26d3422188e82d7d2ccef7e08543d7f2c2b096 :grinning:

Thaodan:
Fixed.

sledges:
Here's my take on the build all solution:

```diff
diff --git a/helpers/build_packages.sh b/helpers/build_packages.sh
index c47edfb..758aacb 100755
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -62,6 +62,7 @@ fi
 eval set -- $OPTIONS
 
 if [ "$#" == "1" ]; then
+    BUILDALL=1
     BUILDDHD=1
     BUILDCONFIGS=1
     BUILDMW=1
@@ -117,10 +118,10 @@ if [ "$BUILDCONFIGS" = "1" ]; then
         sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -m sdk-install -R zypper se -i community-adaptation > /dev/null
         ret=$?
         if [ $ret -eq 104 ]; then
-            BUILDALL=y
+            BUILDALL_OVERRIDE=1
             buildmw -u "https://github.com/mer-hybris/community-adaptation.git" \
                     -s rpm/community-adaptation-localbuild.spec || die
-            BUILDALL=n
+            BUILDALL_OVERRIDE=0
         elif [ $ret -ne 0 ]; then
             die "Could not determine if community-adaptation package is available, exiting."
         fi
diff --git a/helpers/util.sh b/helpers/util.sh
index 97ce2fd..3318af8 100755
--- a/helpers/util.sh
+++ b/helpers/util.sh
@@ -33,7 +33,7 @@
 #
 
 PORT_ARCH="${PORT_ARCH:-armv7hl}"
-BUILDALL=n
+BUILDALL=0
 LOG="/dev/null"
 CREATEREPO="createrepo_c"
 ALLOW_UNSIGNED_RPM=""
@@ -139,7 +139,7 @@ buildversion() {
 }
 
 yesnoall() {
-    if [ $BUILDALL = "y" ]; then
+    if [ $BUILDALL = "1" ] || [ $BUILDALL_OVERRIDE = "1" ]; then
         return 0
     fi
     read -r -p "${1:-} [Y/n/all]" REPLY
@@ -149,7 +149,7 @@ yesnoall() {
        true
        ;;
     [aA]*)
-       BUILDALL=y
+       BUILDALL=1
        true
        ;;
     *)
```

Thaodan:
Mine is at #236. It works the similar but only when the option is set to ask during the build of a pkg. 
Should this apply only to --mw builds or to other section to such as --hal?

sledges:
Tested, works, thanks.

Thaodan:
Changed as requested

sledges:
> Should this apply only to --mw builds or to other section to such as --hal?

It always applied to --mw builds only https://github.com/mer-hybris/droid-hal-device/blob/8a0564f1a74cdc71c81725a79a72d745accfe813/helpers/util.sh#L190

Such functionality should not change. It's only the mw packages that the user cannot select which ones to build from cmdline. They can always combine `-d, -c, and -v` switches as needed, but not mw.

PR #226, opened 2019-08-10
[dhd][udev] Make sure sysfs paths exist

elros34:


mlehtima:
Could you adapt the commit message to match the format used in recent commits? Also could you use only one tag in the beginning of the commit message, in this case udev is probably good.

helpers/makeudev
```diff
@@ -52,7 +52,11 @@ while (<>) {
     print "# sys rule\n";
     s{^/sys/}{};
     my ($dev, $key, $mode, $owner, $group) = split;
-    print qq{DEVPATH=="/$dev", RUN+="/bin/chmod $mode /sys/\$devpath/$key", RUN+="/bin/chown $owner:$group /sys/\$devpath/$key"\n} ;
+    if ($dev =~ m{^devices/system/cpu/cpu\*}) {
+        print qq{DEVPATH=="/$dev", ACTION=="online", TEST=="/sys/\$devpath/$key", RUN+="/bin/chmod $mode /sys/\$devpath/$key", RUN+="/bin/chown $owner:$group /sys/\$devpath/$key"\n} ;
+    } else {
+        print qq{DEVPATH=="/$dev", TEST=="/sys/\$devpath/$key", RUN+="/bin/chmod $mode /sys/\$devpath/$key", RUN+="/bin/chown $owner:$group /sys/\$devpath/$key"\n} ;	
+    }
```
sledges: Sorry, was looking in a wrong place earlier, and now realised that after these changes, the chmod and chown for existing sysfs paths are no longer executed, hence those are now with wrong permissions (thanks @mlehtimaki)

A possible rewrite could be to remove the TEST="..." part and do something like `RUN+="/bin/sh -c \"/bin/chmod $mode /sys/\$devpath/$key ||:\""` (untested)

However, I also tested without the TEST="..." bit, and ACTION="online" also causes permissions+ownership not to be set thereafter.
So please recheck if you have the correct permissions and ownership for existing sysfs paths/nodes on your device.

elros34: Sorry, indeed cpu rules does not work. I made mistake probably because on device I tested it chown/chmod for cpu0 are also executed in init*rc. Additionally adding TEST suppress systemd-udevd fail spam.
Entry from uevend.rc
`/sys/devices/system/cpu/cpu* cpufreq/scaling_governor   0664    system  system`
is converted to:

> /bin/chmod 0664 /sys//devices/system/cpu/cpu1/**cpufreq**/**cpufreq**/scaling_governor"  

so path for such a rule is always incorrect with or without this PR.

Another issue which I have notice is that udev online/add events  for "/devices/system/cpu/cpu1" or "/devices/system/cpu/cpu1/cpufreq" are not always generated at boot time so even with correct rules, correct permissions are set only after some time when cpu1 goes offline->online.
This needs deeper investigation.

PR #231, opened 2019-09-09
dhd: Add Source0 for droid_src_build.

mkosola:
[dhd] Add Source0 for droid_src_build. JB#46848

Source0 was moved to main spec file but version is not known in
that phase on promotion checks. Re-introduce it back after
version is known.

Signed-off-by: Matti Kosola <matti.kosola@jolla.com>

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -138,6 +138,7 @@ Version: 	0.0.6
 %if 0%{?_obs_build_project:1}
 Release: 	1
 %if 0%{?droid_src_build:1}
+Source0: droid-hal-%{rpm_device}-%{version}.tar.bz2
 BuildRequires: droid-bin-src-full
 %else
 # The repo sync service on OBS prepares a 'source tarball' of the rpm


sledges:
Why wouldn't we revert https://github.com/mer-hybris/droid-hal-device/commit/67a0761c5eeba346225edd7eac93a83fc365e1c3#diff-c5e57a2b0b8037c2b84f5942e90daa16 ?

mlehtima:
That uses different source package name. Would that original one work or will it create incorrect name?

mkosola:
Because name needs to be droid-hal-%{rpm_device}-%{version}.tar.bz2

mkosola:
Yeah, original does not work.

sledges:
Yet `Name: 		droid-hal-%{rpm_device}` so this would be equivalent to `%{name}-%{rpm_device}`, and Name is not defined inside a conditional block

mkosola:
%{name} includes %{rpm_device} and that is not resolved properly if we use %{name}-%{version}.tar.bz2

PR #233, opened 2019-10-10
dhd: Create /data. Fixes JB#47683

ibaranovomp:


droid-hal-device.inc
```diff
@@ -656,7 +657,6 @@ _remove_cruft   "/fstab.*"          \
                 "/seapp_contexts"   \
                 "/charger"          \
                 "/res"              \
-                "/data"             \
```
ibaranovomp: In fact, I'm not sure that we need to remove the /data cleanup before it creation. Theoretically there can be some garbage.

droid-hal-device.inc
```diff
@@ -815,6 +815,7 @@ fi
 
 %files -f tmp/droid-hal.files
 %defattr(-,root,root,-)
+/data
```
saukko: This should be %dir /data to ensure we only include the directory for this package, right?

droid-hal-device.inc
saukko: I would just remove it just to be on the safe side.

ibaranovomp: Done

droid-hal-device.inc
ibaranovomp: Done

ibaranovomp:
UP

PR #236, opened 2019-10-21
helpers: Don't ask every time a package is build just ask if $BUILDMW=1 

Thaodan:
As explained by  @mlehtima:
>Also it might not be a good idea to completely remove the possibility
>to skip building certa in packages as often porters have to skip
>package because they have to build an older versi n of some package
>manually due to incompatibility with current Sailfish OS version.

Just ask for confirmation if the pkg should be build explicitly
when $BUILD_INDIVIDUAL_ASK is set or not n.

### DIFF
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -135,6 +135,7 @@ if [ "$BUILDCONFIGS" = "1" ]; then
 fi
 
 if [ "$BUILDMW" = "1" ]; then
+
     sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -R -msdk-install ssu domain sales
     sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -R -msdk-install ssu dr sdk
 
@@ -151,6 +152,7 @@ if [ "$BUILDMW" = "1" ]; then
     pushd $ANDROID_ROOT/hybris/mw > /dev/null
 
     if [ "$BUILDMW_REPO" = "" ]; then
+        BUILDMW_ASK=y
         buildmw -u "https://github.com/mer-hybris/libhybris" || die
 
         if [ $android_version_major -ge 8 ]; then
--- a/helpers/util.sh
+++ b/helpers/util.sh
@@ -33,7 +33,6 @@
 #
 
 PORT_ARCH="${PORT_ARCH:-armv7hl}"
-BUILDALL=n
 LOG="/dev/null"
 CREATEREPO="createrepo_c"
 ALLOW_UNSIGNED_RPM=""
@@ -138,8 +137,8 @@ buildversion() {
     cd ../../
 }
 
-yesnoall() {
-    if [ $BUILDALL = "y" ]; then
+buildmwquery() {
+    if [ ! "$BUILDMW_ASK" = "n" ]; then
         return 0
     fi
     read -r -p "${1:-} [Y/n/all]" REPLY
@@ -149,7 +148,7 @@ yesnoall() {
        true
        ;;
     [aA]*)
-       BUILDALL=y
+       BUILDMW_ASK=n
        true
        ;;
     *)
@@ -187,7 +186,7 @@ buildmw() {
 
     PKG="$(basename ${GIT_URL%.git})"
 
-    if yesnoall "Build $PKG?" ; then
+    if buildmwquery "Build $PKG?" ; then
         # Remove this warning when ngfd-plugin-droid-vibrator will get rid of CMake
         if [ "$GIT_URL" = "ngfd-plugin-droid-vibrator" ]; then
             merror "WARNING: ngfd-plugin-droid-vibrator build is known to halt under various scenarios!"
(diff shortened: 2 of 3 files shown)


helpers/build_packages.sh
```diff
@@ -194,6 +196,7 @@ if [ "$BUILDMW" = "1" ]; then
             # pull device's bluez4 configs correctly
             sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -m sdk-install -R zypper remove bluez-configs-mer
         fi
+        BUILD_INDIVIDUAL_ASK=n
```
sledges: code readability suffers with such approach

mlehtima: I also wondered why not just use BUILD_ALL

Thaodan: I thought it would be better to set that questioning is needed than set when its not needed and renamed the variable to reflect that.

sledges:
Commit's a20e1a42c1dfdbafc2db84d30f4f6198fda3b110 equivalent is already in master, please remove

helpers/build_packages.sh
sledges: I don't think setting this new flag to =y and to =n is done in the right places. Will it ask or not, in cases of:
- `build_packages.sh`
- `build_packages.sh --mw`
- `build_packages.sh --mw=libhybris`

?

helpers/util.sh
```diff
@@ -149,7 +149,7 @@ yesnoall() {
        true
        ;;
     [aA]*)
-       BUILDALL=y
```
sledges: This means BUILDALL (declared at the top of the file) is no longer used anywhere anymore

helpers/util.sh
```diff
@@ -139,7 +139,7 @@ buildversion() {
 }
 
 yesnoall() {
-    if [ $BUILDALL = "y" ]; then
+    if [ -n "$BUILD_INDIVIDUAL_ASK" ] && [ ! $BUILD_INDIVIDUAL_ASK = "n" ]; then
```
sledges: This currently fails shellcheck with SC2086.

It could be rewritten altogether as
```bash
if [ ! "$BUILD_INDIVIDUAL_ASK" = "n" ]; then
```

PS: is the following not preffered/the POSIX way? IMHO, it's easier to notice `!=` than `! ... =`
```bash
if [ "$BUILD_INDIVIDUAL_ASK" != "n" ]; then

**Comment by mlehtima** [comment]
There is no such option as --mw=1 which is mentioned in commit and PR titles. The option --mw has either no parameter or takes repo url or name as parameter.

File: helpers/util.sh (hunk 3 above)
Comment (Thaodan) [reply helpers/util.sh]: Changed as requested.

File: helpers/build_packages.sh (hunk 1 above)
Comment (Thaodan) [reply helpers/build_packages.sh]: The idea was that when $BUILDMW is 1 that the script asks if each pkg should be build, in every other case it doesn't.

**Comment by Thaodan** [comment]
Changed the PR and the commit message to explain the change better.

File: helpers/build_packages.sh (hunk 1 above)
Comment (sledges) [reply helpers/build_packages.sh]: Why is the flag set to =n (`BUILD_INDIVIDUAL_ASK=n`) in such place where it's never read by any line of code again after that?

**Comment by sledges** [comment]
Alternative approach: https://github.com/mer-hybris/droid-hal-device/pull/237

File: helpers/build_packages.sh (hunk 1 above)
Comment (Thaodan) [reply helpers/build_packages.sh]: Because I oversaw that and made an error. Fixed. Thanks for pointing out.

**Comment by Thaodan** [comment]
I fixed the error as pointed out above. I'm open to your approach to however I think quiet parameter for yesnoll/buildmwquery is redundant as it doesn't ask anyway when the option to ask isn't set.

**Comment by Thaodan** [comment]
@sledges I adopted your naming scheme from #237 as it makes more sense.
Also I saw that I set $BUILDMW_ASK a line two high and not after if [  "$BUILDMW_REPO" = "" ] ... and fixed that.

**Comment by sledges** [comment]
> I fixed the error as pointed out above. I'm open to your approach to however I think quiet parameter for yesnoll/buildmwquery is redundant as it doesn't ask anyway when the option to ask isn't set.

But when the option to ask is set (in `build_packages.sh --mw=package` and in `build_packages.sh --config --mw` cases) we still should not ask during single-mw-package builds and a build of community-adaptation during configs respectively.

File: helpers/build_packages.sh [hunk 4]
```diff
@@ -151,6 +152,7 @@ if [ "$BUILDMW" = "1" ]; then
     pushd $ANDROID_ROOT/hybris/mw > /dev/null
 
     if [ "$BUILDMW_REPO" = "" ]; then
+        BUILDMW_ASK=y
```
sledges: This will also ask when launching `build_packages.sh` without any cmdline parameters.
But the goal is to have an unassisted non-interactive automated build when no parameters are given (the first step towards that was done here: https://github.com/mer-hybris/droid-hal-device/pull/149/commits/41f9f2a49be9ce75ec15433da121018017a6ee88)

And that's what https://github.com/mer-hybris/droid-hal-device/pull/237 does

Thaodan: Ah thanks for the explanation, that makes sense. I oversaw that there's no way for BUILDMW=1 && BUILD_REPO="" to know that the script is called to build everything.

PR #237, opened 2019-10-21
helpers: ask to build each package only when --mw is in cmdline

sledges:
This means no questions will be asked during a build invoked without cmdline
switches.


sledges:
@Thaodan, thanks for reminding me about BUILDALL leftover in build_bootimg_packages.sh, I pushed here as well https://github.com/mer-hybris/droid-hal-device/pull/236/files#diff-bc9ec8185b96479a67d264906cd4f4ebL12

PR #238, opened 2019-10-22
helpers: build mw packages if they are in manifest.xml

sledges:
Check for manifest.xml under $ANDROID_ROOT/.repo (or in parent dirs) and build mw packages from there, if at least one path is under hybris/mw.

### DIFF
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -153,7 +153,65 @@ if [ "$BUILDMW" = "1" ]; then
 
     pushd $ANDROID_ROOT/hybris/mw > /dev/null
 
-    if [ "$BUILDMW_REPO" = "" ]; then
+    manifest_lookup=$ANDROID_ROOT
+    while true; do
+        manifest=$manifest_lookup/.repo/manifest.xml
+        if [ -f "$manifest" ] || [ "$manifest_lookup" = "$(dirname "$manifest_lookup")" ]; then
+            break
+        fi
+        manifest=""
+        manifest_lookup=$(dirname "$manifest_lookup")
+    done
+
+    if [ ! "$BUILDMW_REPO" = "" ]; then
+        # No point in asking when only one mw package is being built
+        BUILDMW_QUIET=1
+        if [ -z "$BUILDSPEC_FILE" ]; then
+            buildmw -u "$BUILDMW_REPO" || die
+        else
+            # Supply all given spec files from $BUILDSPEC_FILE array prefixed with "-s"
+            buildmw -u "$BUILDMW_REPO" "${BUILDSPEC_FILE[@]/#/-s }" || die
+        fi
+    elif [ -n "$manifest" ] &&
+         grep -ql hybris/mw $manifest; then
+        buildmw_cmds=()
+        bifs=$IFS
+        while IFS= read -r line; do
+            if [[ $line = *"hybris/mw"* ]]; then
+                IFS="= "$'\t'
+                for tok in $line; do
+                    word=$(echo "$tok" | cut -d \" -f2 | cut -d \' -f2)
+                    if [ "$preword" = "path" ]; then
+                        if [ "$(basename $(dirname "$word"))" = "mw" ]; then
+                            # Only build first level projects
+                            mw=$(basename "$word")
+                        fi
+                    elif [ "$preword" = "spec" ]; then
+                        spec=$word
+                    fi
+                    preword=$word
+                done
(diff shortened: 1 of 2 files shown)


helpers/build_packages.sh
```diff
@@ -151,7 +151,60 @@ if [ "$BUILDMW" = "1" ]; then
 
     pushd $ANDROID_ROOT/hybris/mw > /dev/null
 
-    if [ "$BUILDMW_REPO" = "" ]; then
+    MANIFEST_LOOKUP=$ANDROID_ROOT
```
Thaodan: Paths like this should be quoted or they brake when they contain spaces.

helpers/build_packages.sh
```diff
@@ -151,7 +151,60 @@ if [ "$BUILDMW" = "1" ]; then
...
+    while true; do
+        MANIFEST=$MANIFEST_LOOKUP/.repo/manifest.xml
+        if [ -f $MANIFEST -o "MANIFEST_LOOKUP" == "$(dirname $MANIFEST_LOOKUP)" ]; then
```
Thaodan: Please see: https://github.com/koalaman/shellcheck/wiki/SC2166
Test -o should be avoided. Either use [ expr ] || [ expr ] or [[ expr || expr ]].
Also  == is bash specific which I removed previously in 59a18beb5290876449f00380fcc45fa709bcd635.

helpers/build_packages.sh
```diff
@@ -151,7 +151,60 @@ if [ "$BUILDMW" = "1" ]; then
...
+                        BUILDMW_CMDS+=("$mw")
+                    else
+                        BUILDMW_CMDS+=("$mw:$spec")
+                        spec=
+                    fi
+                    mw=
+                fi
+            fi
+        done < "$MANIFEST"
+        for bcmd in ${BUILDMW_CMDS[@]}; do
```
Thaodan: This should be quoted to prevent word splitting like with "$@".

helpers/build_packages.sh
```diff
@@ -151,7 +151,60 @@ if [ "$BUILDMW" = "1" ]; then
...
+        # No point in asking when only one mw package is being built
+        BUILDMW_QUIET=1
+        if [ -z "$BUILDSPEC_FILE" ]; then
+            buildmw -u $BUILDMW_REPO || die
+        else
+            # Supply all given spec files from $BUILDSPEC_FILE array prefixed with "-s"
+            buildmw -u $BUILDMW_REPO "${BUILDSPEC_FILE[@]/#/-s }" || die
+        fi
+    elif [ -n $MANIFEST ] &&
+       [ -n "$(grep -l hybris/mw $MANIFEST)" ]; then
```
Thaodan: Can you just use the result of grep by doing something like this?:
```
if grep -ql expr $file 
then
..
fi
``` 
This would avoid a sub shell and the need for test.

helpers/build_packages.sh
sledges: That's true for the cases below, but in this simple assignment case it should work without quotes, shellcheck doesn't complain about that (or could you show a case where it breaks?)

```bash
mkdir "space space"
spc=$(ls -d space*)
echo $spc
assigning=$spc
echo $assigning
```

helpers/build_packages.sh
```diff
@@ -151,7 +151,60 @@ if [ "$BUILDMW" = "1" ]; then
 
     pushd $ANDROID_ROOT/hybris/mw > /dev/null
 
-    if [ "$BUILDMW_REPO" = "" ]; then
+    MANIFEST_LOOKUP=$ANDROID_ROOT
+    while true; do
+        MANIFEST=$MANIFEST_LOOKUP/.repo/manifest.xml
+        if [ -f "$MANIFEST" ] || [ "$MANIFEST_LOOKUP" == "$(dirname "$MANIFEST_LOOKUP")" ]; then
```
Thaodan: If you don't use bash features you can still use [ var = var ] here.
Also dirname can be replaced with ${var%/*}, the intentions isn't that obvious this way but it avoid external commands and command substitution.

helpers/build_packages.sh
Thaodan: Ah yes you're right. Shell quoting rules are weird sometimes.

helpers/build_packages.sh
sledges: Looks ok but reduces readability. There are lots of other shell substitutions, so let's leave it at that.

sledges:
Merging today if no objections

PR #239, opened 2019-10-22
helpers: Don't use deprecated std c++0x

Thaodan:
The standard flag c++0x is depreacted in gcc, use recommended c++11 or
gnu++11. This also fixes android 10 builds because libsparse is converted to
c++ typeof starting with android 10/platform-tools 29 and use c
typeof (which is a gnu extension).

### DIFF
--- a/helpers/simg2img.mk
+++ b/helpers/simg2img.mk
@@ -2,7 +2,7 @@
 
 SIMG2IMG_SOURCES?= nosourcessupplied
 
-CXXFLAGS+= -std=c++0x
+CXXFLAGS+= -std=gnu++11
 
 CPPFLAGS+= -Iinclude
 CPPFLAGS+= -I../base/include
--- a/helpers/img2simg.mk
+++ b/helpers/img2simg.mk
@@ -2,7 +2,7 @@
 
 IMG2SIMG_SOURCES?= nosourcessupplied
 
-CXXFLAGS+= -std=c++0x
+CXXFLAGS+= -std=gnu++11
 
 CPPFLAGS+= -Iinclude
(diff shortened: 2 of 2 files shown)


sledges:
Will this work with older gcc's from ancient Android versions (such as kitkat)?

Thaodan:
Which one? It worked with gcc 4.9 for example.

sledges:
Ah they are built in PLATFORM_SDK, not HABUILD_SDK, so it should be fine unless a very old sfos version is used (which a safe bet not happen as we have a cap on the 3.0.0+ version in HADK)

Thaodan:
First version supporting this flag should be gcc 4.7:
https://gcc.gnu.org/projects/cxx-status.html
https://gcc.gnu.org/gcc-4.7/cxx0x_status.html

helpers/simg2img.mk
```diff
@@ -2,7 +2,7 @@
 
 SIMG2IMG_SOURCES?= nosourcessupplied
 
-CXXFLAGS+= -std=c++0x
+CXXFLAGS+= -std=c++11
```
sledges, line 5: We should unify flags to either both gnu+11 or both c++11.

sledges (changes requested):

sledges (changes requested):
nitpick: typo: deprecated (please replace depreacted occurrences with deprecated)

helpers/simg2img.mk
Thaodan: Then it should be gnu++11 as we use typeof.

sledges (approved):

PR #244, opened 2019-10-27
helpers: add building image with mic

sledges:


mlehtima:
A possible future addition could be allowing usage of user defined existing .ks file for mic build. I for example use a modified .ks file for certain devices.

PR #246, opened 2019-11-01
Simplify droid-hal user and group creation

krnlyng:
@ballock I couldn't add you as a reviewer here but please do provide input if you have some.

Only tested on 1 device so far.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -420,8 +420,7 @@ make ANDROID_ROOT=$ANDROID_ROOT IMG2SIMG_SOURCES="$IMG2SIMG_SOURCES" SIMG2IMG_SO
 popd
 
 echo Building uid scripts
-%{dhd_path}/helpers/usergroupgen add > %{dhd_path}/helpers/droid-user-add.sh
-%{dhd_path}/helpers/usergroupgen remove > %{dhd_path}/helpers/droid-user-remove.sh
+%{dhd_path}/helpers/usergroupgen > %{dhd_path}/helpers/droid.ids
 
 echo Building udev rules
 mkdir tmp/udev.rules
@@ -609,13 +608,8 @@ else
 cp -a %{android_root}/hybris/hybris-boot/hb/init-script $RPM_BUILD_ROOT/init-debug
 fi
 
-# droid user support This may be better done by passing a list of
-# users/groups and running 'ensure_usergroups_exist newlist oldlist'
-# which would preserve oldlist in %%post and delete any users no longer
-# needed (unlikely!). This avoids the transient removal of uids and
-# group issues
-install -D %{dhd_path}/helpers/droid-user-add.sh $RPM_BUILD_ROOT%{_libdir}/droid/droid-user-add.sh
-install -D %{dhd_path}/helpers/droid-user-remove.sh $RPM_BUILD_ROOT%{_libdir}/droid/droid-user-remove.sh
+# droid user support
+install -D %{dhd_path}/helpers/droid.ids $RPM_BUILD_ROOT%{_libdir}/droid/droid.ids
 
 # Remove cruft, unless the file is listed in straggler_files
 
@@ -678,7 +672,7 @@ done
 cp tmp/hw-release %{buildroot}/%{_libdir}/droid-devel/hw-release.vars
 
 # This ghost file must exist in the installroot
-touch $RPM_BUILD_ROOT/%{_libdir}/droid/droid-user-remove.sh.installed
+touch $RPM_BUILD_ROOT/%{_libdir}/droid/droid.ids.created
 
 echo "#!/bin/sh" > $RPM_BUILD_ROOT/%{_libdir}/droid/generated-post-scripts.sh
 if [ $ANDROID_VERSION_MAJOR -ge "5" ] && \
@@ -743,7 +737,26 @@ done
 # Only run this during final cleanup
 if [ $1 -eq 0 ]; then
     echo purging old droid users and groups
-    %{_libdir}/droid/droid-user-remove.sh.installed || :
+    cat %{_libdir}/droid/droid.ids.created |while read name id; do
+        userdel -f $name
+        groupdel $name
+        sed -i "/^$name $id\$/d" droid.ids.created
+    done || :
+fi
+
+%pre
+#Compatibility with old droid-hal version - convert previous format
+#of droid-specific user/group information to current one.
+cd %{_libdir}/droid
+if [ $1 -gt 1 -a \
+     -e droid-user-remove.sh.installed -a \
+     -e droid-user-add.sh ]; then
+  #We've got the previous data in a shell-style generated text file. Convert.
+  #Only user/group name and ID is important, username=groupname, uid=gid.
+  sed -n -e 's/^groupadd -g \(.*\) \(.*\)/\2 \1/p' \
+      droid-user-add.sh > droid.ids.created || :
+  #Truncate the old file. rpm will remove it on its own.
+  echo '#!/bin/sh' > droid-user-remove.sh.installed
 fi
 
 %post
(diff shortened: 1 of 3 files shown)


droid-hal-device.inc
```diff
@@ -829,20 +788,100 @@ fi
...
+  #We've got the previous data in a shell-style generated text file. Convert.
+  #Only user/group name and ID is important, username=groupname, uid=gid.
+  sed -n -e 's/^groupadd -g \(.*\) \(.*\)/\2 \1/p' \
+      droid-user-add.sh > droid.ids.created || :
+  #Truncate the old file. rpm will remove it on its own.
+  echo '#!/bin/sh' > droid-user-remove.sh.installed
+fi
+
+%preun users
+# Only run this during final cleanup
```
jusa: I wouldn't delete groups ever. Only if there is a group with differing group id from old one, then we could do

`groupmod --gid <new group id> <group name>`

Deleting groups was the problem to start with, trying to kludge with this is difficult. As the important thing is that whatever the used group array currently active has correct guids better to just update those if they change and leave possibly unused ones as is. This way for example the workaround below for system mode PulseAudio would be unneeded. As pulse was added at some point to audio group, audio group will continue existing and if the guid changes group members are intact.

droid-hal-device.inc
```diff
@@ -829,20 +785,67 @@ fi
...
+
+%post users
+cd %{_libdir}/droid
+
+# NOTE: We do not delete groups/users in any case since that would cause issues
+# with users being removed from groups where they should really be in.
+
+# Now for both install/update add the users and mark which were added already
+echo creating droid users and groups
+# On fresh installs, the file is not there at all and would break grep below
```
jusa: This comment and one above is outdated after previous update.

droid-hal-device.inc
```diff
@@ -829,20 +785,67 @@ fi
...
+DROID_USERS=$(cat droid.ids)
+if [ "$DROID_USERS" != "" ]; then
+    echo "$DROID_USERS" |while read name id; do
```
jusa: This could be simplified to just
`[ -e droid.ids ] && cat droid.ids | while read name id; do`

droid-hal-device.inc
```diff
@@ -829,20 +785,62 @@ fi
...
+# These are the default android ones:
+/usr/bin/groupadd-user audio || :
+/usr/bin/groupadd-user graphics || :
+/usr/bin/groupadd-user system || :
+/usr/bin/groupadd-user input || :
+/usr/bin/groupadd-user camera || :
+/usr/bin/groupadd-user media || :
+/usr/bin/groupadd-user mtp || :
+/usr/bin/groupadd-user bluetooth || :
+/usr/bin/groupadd-user media_rw || :
```
rainemak: Please do not use this oneshot thingie. Just use usermod directly.

rainemak: If just possible... This may break if we have multiple users

droid-hal-device.inc
```diff
@@ -829,20 +785,62 @@ fi
...
+        echo "User $name already existed"
+        usermod -s /sbin/nologin -d / -u $id -g $id $name || :
+    fi
+done
+
+# Now ensure default user has access to various subsystems this HA provides
+# These are the default android ones:
+/usr/bin/groupadd-user audio || :
+/usr/bin/groupadd-user graphics || :
+/usr/bin/groupadd-user system || :
```
rainemak: I don't think that default user should be added system

krnlyng: > I don't think that default user should be added system

That should be filed as a separate bug.

rainemak: Thanks

droid-hal-device.inc
```diff
@@ -829,20 +785,62 @@ fi
...
+echo creating droid users and groups
+[ -e droid.ids ] && cat droid.ids | while read name id; do
+    if ! getent group $name; then
+        echo "Group $name did not exist yet"
+        groupadd -g $id $name || :
+    else
+        echo "Group $name already existed"
+        groupmod -g $id $name || :
+    fi
+    if ! getent passwd $name; then
```
ballock: With the current shape this PR makes a regression where it's no longer supported for the android adaptation to change names and/or IDs of android user/groups.

The pre-PR code handled this by removing all users/groups it created, including the group members. My code tried to avoid the situation by assuming that the package is never removed, just updated. In that case, the users/group whose IDs did not change were not touched, and thus the membership added by other packages was retained. At the same time if an android ID was modified, the change was propagated, and the user/group was removed and re-created with a proper ID. Since it was removed, if any other packages added members to that renumbered group, the membership was lost.

The modified code makes a convenient assumption that we do not need to remove users/groups when the package is removed. After all, some stray user/group that no longer exists in the new Android adaptation should not cause us any issues by staying in the system. The only case where I see it might is if it's got an ID that is re-used in a different Android user/group or in Sailfish.

So, the real problem I see here is if the android IDs change:
- if the name is changed and ID stays the same, we either have 2 user/groups with the same ID or usermod below fails (I didn't check if it would).
- if the ID changes and the name stays the same, this getent will be glad to find the user, and will not add the new, correct one.

My gut feeling tells me that the android IDs should rarely change, if at all. Additions of new entries is supported, so even if there's a new one, it's ok. At the time I was writing the PR I was persuaded that android ID changes actually happen. Still, I think that was probably during early device adaptation process, and I don't think we bumped the android adaptation layer in any device so significantly as to change android IDs. 

Following my gut feeling, the most reasonable thing to do is to have the package create the users/groups and never touch them afterwards. When there's a change in the android adaptation layer, one needs to verify what change it is, and if there's any ID removal or rename, handling for that case needs to be added to the particular device's droid-hal-$device-users instead of the common, generic package that would then only handle additions.

jusa: Comment on the problems you mentioned:
1. If the name is changed and ID stays the same, default useradd/groupadd will not allow duplicate gid, but it can be forced. Though I'd argue we shouldn't have duplicate gids.
2. If the ID changes and the name stays the same, current PR would change the uid/gid to new one, so that's not a problem.

From those issues first one is something that *could* be an issue. If so, for that case it might make sense to implement something where the old group is removed, new one created and users from previous group are re-added to the new one.

krnlyng: 2) is a problem if the group is shared by SFOS and files in the filesystem already exist under that group, i.e. after changing the ID things would loose access to those files
we had this case already once with the dbus group, (android uses a different id for dbus) so we had to rename the android dbus group to adbus.

1) i think @jusa's suggestion for handling this issue could be done but that would also be handled specially if the group is shared with SFOS, since we should not remove SFOS groups at all.

droid-hal-device.inc
```diff
@@ -196,6 +198,17 @@ Summary: Development files for droid-hal device: %{rpm_device}
 Device specific droid headers for %{rpm_device}.
 Needed by libhybris
 
+################
+%package users
+Group:	Tools
+Requires: %{name} = %{version}-%{release}
+Provides: droid-hal-users
+Summary: Create users and groups for device: %{rpm_device}
```
jusa: nit: whitespace in the package definition

droid-hal-device.inc
jusa: > With the current shape this PR makes a regression where it's no longer supported for the android adaptation to change names and/or IDs of android user/groups.

On the contrary, with the change group names and/or gids are updated to the new ones if there are changes. However, like you said, if a group is not used in new version it is not removed but having unused groups is less problematic than removing them.

ballock: Indeed, I must have overlooked the else statement, that's why my understanding of the code was different from what happens as per @jusa's point 2.

With the updates by @krnlyng, I think the right thing, "best effort" happens. If we find a duplicate ID, we rename the group, if we find a duplicate name, we renumber that group.

PR #252, opened 2019-11-21
dhd: Users package post uses oneshot.

jusa:


droid-hal-device.inc
```diff
@@ -200,9 +200,9 @@ Needed by libhybris
 ################
 %package users
 Group: Tools
-Requires: %{name} = %{version}-%{release}
 Provides: hardware-adaptation-setup
 Summary: Create users and groups for device: %{rpm_device}
+%{_oneshot_requires_post}
```
mkosola: I think this should be:
Requires: oneshot
%{_oneshot_requires_post}

jusa: Macro already adds dependency to oneshot.

PR #254, opened 2019-11-25
helpers: Fix droidmedia and audioflingerglue local builds in online mode

mlehtima:
[helpers] Fix droidmedia and audioflingerglue local builds in online mode

### DIFF
--- a/helpers/util.sh
+++ b/helpers/util.sh
@@ -218,7 +218,7 @@ buildmw() {
             fi
 
             pushd $PKG > /dev/null || die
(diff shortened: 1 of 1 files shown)


JamiKettunen:
I already mentioned this on the porters IRC too, but the check `"$PKG" != "audioflingerglue-localbuild" && "$PKG" != "droidmedia-localbuild"` could be shortened to `"$PKG" != *"-localbuild"`.
Any comment @sledges?

Thaodan:
if you want to keep using posix test use [ ] && [ ]

PR #258, opened 2019-12-18
Store user required groups instead of adding immediately.

jusa:
Instead of adding default user to hardware adaptation required groups
during package install store the groups to filesystem for later access.
Then when new user is created this list can be consulted for the correct
groups the user needs to be in.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -202,7 +202,6 @@ Needed by libhybris
 Group: Tools
 Provides: hardware-adaptation-setup
 Summary: Create users and groups for device: %{rpm_device}
-%{_oneshot_groupadd_requires_post}
 
 %description users
 This package provides android users and groups which are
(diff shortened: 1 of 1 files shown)


droid-hal-device.inc
```diff
@@ -691,14 +690,34 @@ done
 # for use in the -devel package
 cp tmp/hw-release %{buildroot}/%{_libdir}/droid-devel/hw-release.vars
 
-echo "#!/bin/sh" > $RPM_BUILD_ROOT/%{_libdir}/droid/generated-post-scripts.sh
+# Store the groups that users need to be in to access various subsystems this HA provides
+# These are the default android ones:
+
+mkdir -p %{buildroot}/%{_libdir}/hw-group.d
```
Tomin1: missing droid directory from the path?

jusa: Yep. Good catch, updated.

PR #260, opened 2019-12-22
dhd: Fix Android 6 <= based builds

AndreySV:
this problem happens if Android version is <= 6 and there is file
hybris_extras.rc in system/etc/init. It is introduced by commit
1fb8758dafa5 ("dhd: add support for one extra init rc, which can be
used to provide info from the adaptation to sfos")

In this case build_packages.sh fails with following error:

$ rpm/dhd/helpers/build_packages.sh

+ '[' 4 -ge 7 ']'
+ echo ''
+ '[' -f ./out/target/product/hammerhead/system/etc/init/hybris_extras.rc ']'
+ cp -a ./out/target/product/hammerhead/system/etc/init/hybris_extras.rc /media/data/port/android/installroot/usr/libexec/droid-hybris/system/etc/init/
cp: cannot create regular file `/media/data/port/android/installroot/usr/libexec/droid-hybris/system/etc/init/': No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.SrHvD1 (%install)

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -533,13 +533,14 @@ if [ $ANDROID_VERSION_MAJOR -ge "7" ]; then
 mkdir -p $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system/etc/init
 cp -a %{android_root}/out/target/product/%{device}/system/etc/init/servicemanager.rc $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system/etc/init/
 echo "%{_libexecdir}/droid-hybris/system/etc/init/servicemanager.rc" > tmp/droid-hal.files
-else
-echo "" > tmp/droid-hal.files
-fi
 
 if [ -f %{android_root}/out/target/product/%{device}/system/etc/init/hybris_extras.rc ]; then
 cp -a %{android_root}/out/target/product/%{device}/system/etc/init/hybris_extras.rc $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system/etc/init/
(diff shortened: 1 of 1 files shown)


droid-hal-device.inc
```diff
@@ -538,6 +538,7 @@ echo "" > tmp/droid-hal.files
 fi
 
 if [ -f %{android_root}/out/target/product/%{device}/system/etc/init/hybris_extras.rc ]; then
```
sledges: If `.../etc/init` should never exist in Android 6 (or earlier) adaptations, why not move this part into the if block just above and re-use the mkdir? ^

AndreySV: As I worked through HADK and built hybris-hal for hammerhead (CyanogenMod 11.0 / Android 4.4). ./out/target/product/hammerhead/system/etc/init/hybris_extras.rc was created.

sledges: @krnlyng, @mlehtima, could you please tell us if there is any part in hybris Android 6 (and earlier) adaptations which would read and/or run the .rc files under `/usr/libexec//droid-hybris/system/etc/init/`?

mlehtima: Nothing reads from that folder in Android 6 or older bases. Hybris version of Android 7 base i.e. hybris-14.1 is the first one to use that folder, correct place for fixing the issue is in hybris-boot repo which seems to create the file on all Android base versions. The file is being created after this commit in hybris-boot https://github.com/mer-hybris/hybris-boot/commit/46095a4d7819962acfb49cdf88cb93d7c0a911fc

AndreySV: @sledges, @mlehtima I've updated this PR as you recommended and made PR to fix original source of problem in hybris-boot (https://github.com/mer-hybris/hybris-boot/pull/184)

PR #278, opened 2020-09-09
Add missing packages, fix variable scoping bug

moralrecordings:
This patch fixes the following:

- Prevent the value of DO_NOT_INSTALL from leaking across calls to deploy()
- Add nfcd and libdbuslog packages, as dependencies for nfcd-binder-plugin
- Add pulseaudio-module-keepalive, as a dependency for pulseaudio-modules-droid


0312birdzhang:
> 
>     * Add nfcd and libdbuslog packages, as dependencies for nfcd-binder-plugin
> 

Not all devices have nfc.

moralrecordings:
Okay, should that mean nfcd-binder-plugin doesn't belong on the list? As of now, if you try and run build_packages.sh in a clean environment, the process will crash because it can't find a bleeding-edge copy of nfcd in the standard repos. Is there any harm to building a package for nfcd/nfcd-binder-plugin if there isn't nfc support on the device.

PR #280, opened 2020-10-19
dhd: Fix android-config.h generating with busybox sed

Thaodan:


droid-hal-device.inc
```diff
@@ -596,7 +596,8 @@ popd 1>/dev/null
 
 # Add our config into the default android-config.h
 echo Making new $HDRS/android-config.h
-sed '/CONFIG GOES HERE/,$d' $HDRS/android-config.h > $HDRS/android-config.h.new
+TOKEN_LINE=$(sed -n '/CONFIG GOES HERE/=' $HDRS/android-config.h)
```
sledges, line 599: Uppercase variables are usually for shell's use, or conventionally for global exports. Script-internal (local) variables best look when in lowercase.

sledges (changes requested):
Commit should read ```android-config.h``` I believe.

Thaodan:
That is true for shell scripts but in this case all variables are mostly 
uppercase and wanted ti follow the rest of the specfile.

droid-hal-device.inc
sledges: > That is true for shell scripts but in this case all variables are mostly uppercase and wanted ti follow the rest of the specfile.

That doesn't mean we shouldn't strive to improve our coding style:)

sledges:
> Commit should read `android-config.h` I believe.

As well as the next 3 commits do not seem to belong to this PR topic.

sledges:
> As well as the next 2 commits do not seem to belong to this PR topic.

I have included those two useful commits in https://github.com/mer-hybris/droid-hal-device/pull/282

droid-hal-device.inc
```diff
@@ -373,10 +373,10 @@ fi
 %if 0%{?_obs_build_project:1}
 
 # Set up kernel extra version for OBS kernel builds
-if EXTRAVERSION_OLD=$(cat %android_root/kernel/Makefile | grep "EXTRAVERSION ="); then
-    sed -i "s/$EXTRAVERSION_OLD/EXTRAVERSION = +%{version}/g" %android_root/kernel/Makefile
-elif EXTRAVERSION_OLD=$(cat %android_root/linux/kernel/Makefile | grep "EXTRAVERSION ="); then
-    sed -i "s/$EXTRAVERSION_OLD/EXTRAVERSION = +%{version}/g" %android_root/linux/kernel/Makefile
+if extra_version_old=$(grep"EXTRAVERSION =" %android_root/kernel/Makefile); then
```
sledges, line 376: ```extra_version_old``` - no such variable referenced afterwards
```grep"EXTRAVERSION ="``` - space missing (will fail as command)
Have you tested this .spec in action?

**Review by sledges — CHANGES_REQUESTED** [review]

**Comment by sledges** [comment]
> Commit should read `android-config.h` I believe.

Commit still references a non-existing `droid-config.h` file.

File: droid-hal-device.inc (hunk 2 above)
Comment (Thaodan) [reply droid-hal-device.inc]: Yes I had after I fixed the typo.

File: droid-hal-device.inc [hunk 3]
```diff
@@ -373,10 +373,10 @@ fi
 %if 0%{?_obs_build_project:1}
 
 # Set up kernel extra version for OBS kernel builds
-if EXTRAVERSION_OLD=$(cat %android_root/kernel/Makefile | grep "EXTRAVERSION ="); then
-    sed -i "s/$EXTRAVERSION_OLD/EXTRAVERSION = +%{version}/g" %android_root/kernel/Makefile
-elif EXTRAVERSION_OLD=$(cat %android_root/linux/kernel/Makefile | grep "EXTRAVERSION ="); then
-    sed -i "s/$EXTRAVERSION_OLD/EXTRAVERSION = +%{version}/g" %android_root/linux/kernel/Makefile
+if extra_version_old=$(grep "EXTRAVERSION =" %android_root/kernel/Makefile); then
```
sledges, line 376: This is still not fixed, should be `extraversion_old`.

sledges (changes requested):

droid-hal-device.inc
```diff
@@ -421,7 +421,7 @@ fi
 echo Building local tools
```
sledges, line 421: There are 4 more occurrences of `test` still not converted to `[ ]`. 3 above and one way below.

sledges (changes requested):

sledges (changes requested):
Variables still in uppercase:
https://github.com/mer-hybris/droid-hal-device/blob/0a2b1ef02429c86a9bf81fd51fec6a3afc71d413/droid-hal-device.inc#L408
https://github.com/mer-hybris/droid-hal-device/blob/0a2b1ef02429c86a9bf81fd51fec6a3afc71d413/droid-hal-device.inc#L697

sledges (approved):

PR #282, opened 2020-10-21
Improve package version handling

sledges:
@Thaodan implemented ability to instruct mb2 to skip generating version from git tag when such is not available (droidmedia-localbuild) or invalid, this way getting rid of couple of old hacks.

Another feature is to recover tags from shallow clones or if working offline, inform the user about it.

### DIFF
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -49,6 +49,8 @@ Usage: $0 [OPTION]..."
    -D, --do-not-install
                    useful when package is needed only in the final image
                    especially when it conflicts in an SDK target
+   -N  --no-auto-version
+                   Tell mb2 to not fix the version inside a spec file
    -o, --offline   build offline after all repos have been cloned or refreshed
    -n, --no-delete do not delete existing packages when adding to repo
 
@@ -57,7 +59,7 @@ EOF
     exit 1
 }
 
-OPTIONS=$(getopt -o hdcm::gvib:s:Don -l help,droid-hal,configs,mw::,gg,version,mic,build:,spec:,do-not-install,offline,no-delete -- "$@")
+OPTIONS=$(getopt -o hdcm::gvib:s:DonN -l help,droid-hal,configs,mw::,gg,version,mic,build:,spec:,do-not-install,offline,no-delete,no-auto-version -- "$@")
 
 if [ $? -ne 0 ]; then
     echo "getopt error"
@@ -83,6 +85,7 @@ while true; do
       -d|--droid-hal) BUILDDHD=1 ;;
       -c|--configs) BUILDCONFIGS=1 ;;
       -D|--do-not-install) DO_NOT_INSTALL=1;;
+      -N|--no-auto-version) NO_AUTO_VERSION=--no-fix-version ;;
       -m|--mw) BUILDMW=1
           BUILDMW_ASK=1
           case "$2" in
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -314,16 +314,12 @@ if [ "$BUILDGG" = "1" ]; then
         fi
         rpm/dhd/helpers/pack_source_droidmedia-localbuild.sh "$droidmedia_version"
         mkdir -p hybris/mw/droidmedia-localbuild/rpm
-        if [ -d .git ] && [ ! -d hybris/mw/droidmedia-localbuild/.git ]; then
-            # for the top-level .git adaptations such as Xperia 10 Android 9, otherwise mb2 will complain:
-            (cd hybris/mw/droidmedia-localbuild; git init; git commit --allow-empty -m "initial")
-        fi
         cp rpm/dhd/helpers/droidmedia-localbuild.spec hybris/mw/droidmedia-localbuild/rpm/droidmedia.spec
         sed -ie "s/0.0.0/$droidmedia_version/" hybris/mw/droidmedia-localbuild/rpm/droidmedia.spec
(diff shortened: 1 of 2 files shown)


Thaodan:
Thanks for moving the commits here. I'll remove them on my PR.

helpers/util.sh
```diff
@@ -170,18 +209,21 @@ buildmw() {
     #  -b     Branch to use. If none supplied, use default.
     #  -s     .spec file to use. Can be supplied multiple times.
     #         If empty, will use all .spec files from $PKG/rpm/*.
+    #  -N     Tell scratchbox to not fix the version inside a spec file.
 
     local GIT_URL=""
     local GIT_BRANCH=""
     local MW_BUILDSPEC=""
+    local NO_AUTO_VERSION
```
sledges, line 217: @Thaodan, `local` means that `NO_AUTO_VERSION` will not inherit the value if it was passed as cmdline arg to `build_packages.sh`, but being able to rebuild any package without autofixing its version would be a good additional feature.

Thaodan: It will as when you set it before in globally it will take this and put in to the local copy of the variable.

sledges: I've run a minimum test case and it's not picking up the value:
```bash
#!/bin/bash

var=glob

f() {
  local var
  echo $var
}

f
```

Thaodan: Than change it to local NO_AUTO_VERSION=$AUTO_VERSION this works as I said/it should.

sledges: Changed to `local NO_AUTO_VERSION=$NO_AUTO_VERSION` with added comment, otherwise it'd blow code reader's mind:)

helpers/build_packages.sh
```diff
@@ -49,6 +49,8 @@ Usage: $0 [OPTION]..."
    -D, --do-not-install
                    useful when package is needed only in the final image
                    especially when it conflicts in an SDK target
+   -N  --no-auto-version
+                   Tell scratchbox to not fix the version inside a spec file
```
jusa, line 53: This and other places where this is mentioned should have mb2 instead of scratchbox.

helpers/util.sh
```diff
@@ -105,6 +105,45 @@ initlog() {
...
+        die "get_package_version(): not within the $pkg directory"
+    fi
+    version=$(get_last_tag)
+    if [ -z "$version" ]; then
+        if [ "$BUILDOFFLINE" = "1" ]; then
+            die "Could not get version for $pkg: make sure it is cloned without --depth=1 or clone-depth=\"1\""
+        else
+            unshallow_attempt=$(git fetch --unshallow 2>&1)
+            ret=$?
+            if [ $ret -eq 128 ]; then
```
jusa, line 124: Maybe better to check for non-0 return? so $ret -ne 0

helpers/build_packages.sh
```diff
@@ -328,22 +331,21 @@ if [ "$BUILDGG" = "1" ]; then
         minfo "Not building audioflingerglue and pulseaudio-modules-droid-glue due to pulseaudio-modules-droid-hidl in patterns"
     elif grep -qs "- pulseaudio-modules-droid-glue" "$pattern_lookup" ||
          grep -qs "Requires: pulseaudio-modules-droid-glue" "$metapackage_lookup"; then
-        audioflingerglue_version=$(git --git-dir external/audioflingerglue/.git describe --tags 2>/dev/null | sed -r "s/\-/\+/g")
+        pkg=audioflingerglue
+        cd external/$pkg || die "Could not change directory to external/$pkg"
+        audioflingerglue_version=$(get_package_version "$pkg")
         if [ -z "$audioflingerglue_version" ]; then
```
jusa, line 337: get_package_version() already exits on failure.

helpers/build_packages.sh
```diff
@@ -301,23 +304,23 @@ if [ "$BUILDGG" = "1" ]; then
...
     if grep -qs "^- gstreamer1.0-droid" "$pattern_lookup" ||
        grep -qs "^Requires: gstreamer1.0-droid" "$metapackage_lookup"; then
-        droidmedia_version=$(git --git-dir external/droidmedia/.git describe --tags 2>/dev/null | sed -r "s/\-/\+/g")
+        pkg=droidmedia
+        cd external/$pkg || die "Could not change directory to external/$pkg"
+        droidmedia_version=$(get_package_version "$pkg")
         if [ -z "$droidmedia_version" ]; then
-            # in case of shallow clone:
-            droidmedia_version=999.99999999.99
+            # Could not obtain version, function call will have shown the error
```
jusa, line 311: get_package_version() already exits on failure.

jusa (changes requested):
Nothing major, some small things.

helpers/build_packages.sh
sledges: `get_package_version()` is called in a sub-shell, thus exit just exits that sub-shell.

helpers/build_packages.sh
sledges: `get_package_version()` is called in a sub-shell, thus exit just exits that sub-shell.

helpers/build_packages.sh
sledges: Done.

helpers/util.sh
sledges: Done.

helpers/build_packages.sh
jusa: ahh, you're right.

jusa (approved):

PR #287, opened 2020-11-20
helpers: Fix audioflingerglue local build.

mlehtima:
[helpers] Fix audioflingerglue local build.


sledges (approved):
I thought that's no longer used:)

mlehtima:
Older devices can still use it if they have not transitioned to the new way.

PR #288, opened 2020-12-10
helpers: Add option to build MW at specific revisions

JamiKettunen:
The devel repos for many packages can move fast at times and cause breakages when building againt the current latest SFOS release, so having a convenient switch to quickly checkout a specific revision of a middleware component simply makes sense.

This works for both git tags as well as commit hashes.

### DIFF
--- a/helpers/build_packages.sh
+++ b/helpers/build_packages.sh
@@ -44,6 +44,8 @@ Usage: $0 [OPTION]..."
    -v, --version   build droid-hal-version
    -i, --mic       build image
    -b, --build=PKG build one package (PKG can include path)
+   -r, --revision=REV optionally used with -m
+                   build package version from a specific git tag or commit hash
    -s, --spec=SPEC optionally used with -m or -b
                    can be supplied multiple times to build multiple .spec files at once
    -D, --do-not-install
@@ -59,7 +61,7 @@ EOF
     exit 1
 }
 
-OPTIONS=$(getopt -o hdcm::gvib:s:DonN -l help,droid-hal,configs,mw::,gg,version,mic,build:,spec:,do-not-install,offline,no-delete,no-auto-version -- "$@")
+OPTIONS=$(getopt -o hdcm::gvib:r:s:DonN -l help,droid-hal,configs,mw::,gg,version,mic,build:,revision:,spec:,do-not-install,offline,no-delete,no-auto-version -- "$@")
 
 if [ $? -ne 0 ]; then
     echo "getopt error"
@@ -96,6 +98,11 @@ while true; do
       -b|--build) BUILDPKG=1
           case "$2" in
               *) BUILDPKG_PATH=$2;;
+          esac
+          shift;;
+      -r|--revision) BUILDREV=1
+          case "$2" in
+              *) BUILDREV_ARGS="-r $2";;
           esac
           shift;;
       -s|--spec) BUILDSPEC=1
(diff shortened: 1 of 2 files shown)


helpers/build_packages.sh
```diff
@@ -44,6 +44,8 @@ Usage: $0 [OPTION]..."
    -v, --version   build droid-hal-version
    -i, --mic       build image
    -b, --build=PKG build one package (PKG can include path)
+   -r, --revision=REV
```
Thaodan, line 47: Please be more verbose here.

helpers/util.sh
```diff
@@ -264,8 +267,14 @@ buildmw() {
...
             if [[ "$BUILDOFFLINE" = "" && "$PKG" != *"-localbuild" ]]; then
-                minfo "pulling updates..."
-                git pull >>$LOG 2>&1|| die "pulling of updates failed"
+                if [ -z "$GIT_REVISION" ]; then
+                    minfo "pulling updates..."
+                    git pull --recurse-submodules >>$LOG 2>&1|| die "pulling of updates failed"
+                else
+                    minfo "setting version to $GIT_REVISION..."
+                    git fetch >>$LOG 2>&1
+                    git reset --hard $GIT_REVISION --recurse-submodules >>$LOG 2>&1|| die "version reset failed"
```
Thaodan, line 276: Please check if local changes exist before doing any changes on the git repository.
See here for an example how to check for local changes:
https://aur.archlinux.org/cgit/aur.git/tree/build_pkgs?h=linux-pf#n15

Thaodan (changes requested):

JamiKettunen:
Would it make sense to also `git clean -df` when switching revision? I know some building some packages leaves residue files in-tree that could potentially cause issues when building another version of the package.

helpers/util.sh
```diff
@@ -264,8 +267,15 @@ buildmw() {
 
             pushd $PKG > /dev/null || die
             if [[ "$BUILDOFFLINE" = "" && "$PKG" != *"-localbuild" ]]; then
-                minfo "pulling updates..."
-                git pull >>$LOG 2>&1|| die "pulling of updates failed"
+                if [ -z "$GIT_REVISION" ]; then
+                    minfo "pulling updates..."
+                    git pull --recurse-submodules >>$LOG 2>&1|| die "pulling of updates failed"
+                else
+                    git diff-index --quiet HEAD -- || die "workdir of $PKG is not clean, refusing to reset version"
```
Thaodan, line 274: Please be more explicit.
Unclean can be files from previous builds or uncommitted changes which is something else entirely,
I suggest:
```suggestion
                    git diff-index --quiet HEAD -- || die "Package $PKG has  is uncommitted changes, please commit first, refusing to reset version"
```

Thaodan (changes requested):

helpers/util.sh
krnlyng: git uses both synonymously:
nothing to commit, working tree clean

but both ways of writing it (seperately) are ok...

JamiKettunen: @Thaodan opinions? I personally don't care what it says, I'd like to get this merged finally after over a year

Thaodan: ""Package $PKG has uncommitted changes, please commit first, refusing to reset version" sounds good, please adjust then it's good to go IMHO.

JamiKettunen: Done

0312birdzhang:
I would ask when will this pr merged? 
On 5.0.0.29 hybris-14.1 armv7hl base, the lastest `geoclue-providers-hybris` failed to build and the lastest `sensorfw` failed to start.
Really need specific revisions support on MW, not some hacks.

PR #289, opened 2021-01-19
dtbo: allow overriding binary name

sledges:


droid-hal-device.inc
```diff
@@ -53,6 +53,7 @@
 #                      from droid-hal-device
 # bootloader_binary: define bootloader binary name when aboot is packaged
 #                      separately from droid-hal-device
+# dtbo_binary:       override dtbo binary name. Defaults to dtbo.img
```
Thaodan, line 56: I think it should be dtbo_image to fit with the rest.

droid-hal-device.inc
```diff
@@ -95,6 +96,9 @@ such_macro supports spaces\
 %if 0%{!?dhd_path:1}
 %define dhd_path rpm/dhd
 %endif
+%if 0%{!?dtbo_binary:1}
+%define dtbo_binary dtbo.img
```
Thaodan, line 100: ^

droid-hal-device.inc
```diff
@@ -813,7 +817,7 @@ fi
 
 touch dtbo.files
 
-if cp out/target/product/%{device}/dtbo.img $RPM_BUILD_ROOT/boot &> /dev/null; then
+if cp out/target/product/%{device}/%{dtbo_binary} $RPM_BUILD_ROOT/boot/dtbo.img &> /dev/null; then
```
Thaodan, line 820: ^

Thaodan (changes requested):

droid-hal-device.inc
sledges: It's binary, because it fits bootloader_binary, they are both images.

Thaodan:
> > @@ -53,6 +53,7 @@
> 
>  #                      from droid-hal-device
>  # bootloader_binary: define bootloader binary name when aboot is 
packaged
>  #                      separately from droid-hal-device
> +# dtbo_binary:       override dtbo binary name. Defaults to dtbo.img
> 
> It's binary, because it fits bootloader_binary, they are both images.
OK at least AOSP talks about dtbo img.

jusa (approved):

PR #297, opened 2021-02-24
dhd: Move droid headers to /usr/include

mlehtima:
[dhd] Move droid headers to /usr/include

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -526,7 +526,7 @@ mkdir -p $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/lib-dev-alog/
 mkdir -p $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/lib64-dev-alog/
 %endif
 mkdir -p $RPM_BUILD_ROOT%{_libexecdir}/droid-hybris/system
-mkdir -p $RPM_BUILD_ROOT%{dhdlibdir}/droid-devel/
+mkdir -p $RPM_BUILD_ROOT%{_includedir}/droid-devel/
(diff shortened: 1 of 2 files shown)


droid-hal-device.inc
```diff
@@ -764,7 +764,7 @@ done
 #mv $RPM_BUILD_ROOT/charger $RPM_BUILD_ROOT/sbin/droid-hal-charger
 
 # for use in the -devel package
-cp tmp/hw-release %{buildroot}/%{dhdlibdir}/droid-devel/hw-release.vars
+cp tmp/hw-release %{buildroot}/%{_includedir}/droid-devel/hw-release.vars
```
sledges, line 767: Fixes then also required in:
https://github.com/mer-hybris/pulseaudio-modules-droid/blob/22be43a324fee369c488c7b6b12282b133b38b95/rpm/pulseaudio-modules-droid.spec#L48
https://github.com/mer-hybris/droid-hal-device/blob/4b9c268bfa0429fa3e7e0db1c5bf72bba3cc04f1/helpers/build_packages.sh#L208
Also in qt5-qpa-hwcomposer-plugin (this has /usr/lib/ hardcoded, probably not used as aarch64 didn't break)
This is as far as I could find, maybe more exist:)

mlehtima: There is already the fix for pulseaudio-modules-droid which has not been included yet but will be added within the pulseaudio update changes. qt5-qpa-hwcomposer-plugin used to have hardcoded path but I fixed that already during aarch64 fixing. I did check other projects already before and couldn't find anything that would break. I will do test build for some older adaptation to make sure nothing breaks.

sledges (approved):

PR #302, opened 2021-06-11
[dhd] Allow bundling keys into kernel and bootloader. JB#54649 OMP#OS-7115

dseight:


droid-hal-device.inc
```diff
@@ -385,6 +389,25 @@ if (grep -q '^TARGET_ARCH := arm64' %{android_root}/device/*/*/BoardConfig*.mk);
 fi
 %endif
 
+%if 0%{?with_system_keys:1}
+KERNEL_DIRS=$(find %android_root -maxdepth 1 -type d \( -name "kernel-*.*" -o -name "kernel" \) &&
+              find %android_root -maxdepth 2 -type d -path "*/linux/kernel")
+
+# Copy trusted keys into kernel source directory
+echo "$KERNEL_DIRS" |
+while IFS= read -r kernel; do
```
Thaodan, line 398: You can directly pass the find output to read, see: https://github.com/koalaman/shellcheck/wiki/SC2044

droid-hal-device.inc
```diff
@@ -385,6 +389,25 @@ if (grep -q '^TARGET_ARCH := arm64' %{android_root}/device/*/*/BoardConfig*.mk);
...
+    cp -r /etc/keys/kernel/* ${kernel}/
```
Thaodan, line 399: This looks weird how could or should come these keys from the rootfs of the build host?

Thaodan (changes requested):
this looks very vendor specific is that standardized between vendors e.g other socs?

droid-hal-device.inc
dseight: Passing output directly is a bash-specific thing, it won't work with plain POSIX shell. Is it okay to use bashisms here?

droid-hal-device.inc
dseight: They are supposed to be provided by `system-keys-kernel` package.

dseight:
> this looks very vendor specific is that standardized between vendors e.g other socs?

All of the vendors that I've seen so far use the same paths for bootloader. At least, these paths are the same on something like a 15 different device models (most of them were Mediatek-based devices and some are Qualcomm-based devices).

droid-hal-device.inc
Thaodan: Shouldn't they go then into /usr/lib?

dseight: `/usr/lib` is [intended](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s06.html) for storing object files and libraries. But keys are neither of them, they are much closer to configuration files (which [should go](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html) into `/etc/*`). Also, Linux Kernel is using /etc/keys/ path [by default](https://github.com/torvalds/linux/blob/122fa8c588316aacafe7e5a393bb3e875eaf5b25/security/integrity/ima/Kconfig#L308) for storing IMA certificate.

So, `/etc/keys` seems to be a better fit here.

dseight: While I am agree on moving static data to `/usr` (which especially makes sense on RO filesystems), I'm not really sure that `/usr/lib` fits well for storing keys and certificates.

Maybe `/usr/share` then? As `/usr/lib` is primarily targeted for storing executable things (e.g. libraries and plugins), and quite rarely used for storing arbitrary data.

PR #304, opened 2021-07-06
dhd: Don't terminate build if files are missing build ids.

mlehtima:
[dhd] Don't terminate build if files are missing build ids. JB#52914


mkosola (approved):

Thaodan:
Can you add a comment explaining the reason for enabling this in the spec file?

mkosola:
> Can you add a comment explaining the reason for enabling this in the spec file?

Hmm, there is already comment what it is doing

mkosola (approved):

PR #310, opened 2021-10-31
droidmedia-localbuild: Don't package droidmedia-devel by default

Thaodan:
Since droidmedia-devel comes from adaption-common

[droidmedia-localbuild] Don't package droidmedia-devel by default. JB#55878


sledges:
Porter can always rebuild droidmedia-devel from source (modifying it beforehand if necessary):
```
buildmw -u "https://github.com/sailfishos/droidmedia.git" \
        -s rpm/droidmedia-devel.spec
```

PR #314, opened 2021-12-06
Build droidmedia-devel directly from source repo

moralrecordings:
Ever since [this commit](https://github.com/sailfishos/droidmedia/commit/7360103b7a78f5d92cde147680fbe6f8c68b8ea4) to droidmedia in July, local builds have been broken. This is due to gst-droid requiring a pkgconfig file (droidmedia.pc) which isn't built here as part of droidmedia-localbuild.

This patch fixes the `./rpm/dhd/helpers/build_packages.sh --gg` step. It looks like this repo might be in the middle of a refactor that also fixes the issue, but in the meantime if you're trying to just build something locally here's how to sort it out.


Thaodan:
You don't have to build droidmedia-devel at all see:
https://github.com/mer-hybris/droid-hal-device/pull/313

PR #318, opened 2022-01-30
helpers: add --quiet argument for non-interactive MW building

JamiKettunen:
Without this one had to do something like `echo a | build_packages.sh --mw`, but we might as well include a flag that does the same thing.


Thaodan:
I don't think this is a good change since when you pass --quiet it still talks 
too you.
I think the script shouldn't ask in the first place if you want to build all.
There are two choices that  make sense imho, either:
1. Make so that --mw with no package passed assumes to build all packages.
2. Ask to explicitly to pass --mw=all if the user wants to build all and fail 
of if all or package name wasn't passed.

JamiKettunen:
@Thaodan why not a third option: keep current behavior of the script but add the tweaks of this PR behind `--mw=all`? I can see how the interactive option is useful as well for some people due to MW package builds breaking somewhat often, so then that package can be reset back to a known working version, run `--mw` and choose `no` for most except `all` when the previously broken package shows up again, then the script continues to build the remaining MW in order.

sledges:
Here's 4th option: add `-y, --non-interactive` global switch and latch from MW side (later other parts can also take use easily), also the short cmdline gets simpler: `-ym` instead of `-m all`

Thaodan:
That would work, I think the default should be what is used most.
Building all or just building some?
I think building all in that case asking if I really want to do that is annoying.

PR #320, opened 2022-02-09
localbuild: fix build flow with latest SDK (based on 4.3.0.15)

sledges:


helpers/util.sh
```diff
@@ -325,6 +325,7 @@ build() {
     fi
     for SPEC in $SPECS ; do
         minfo "Building $SPEC"
+        mb2 -s $SPEC -t $VENDOR-$DEVICE-$PORT_ARCH $NO_AUTO_VERSION build-requires reset
```
martyone, line 328: Is this still needed with 4.3.0.15-based chroot?

sledges: Yes, it just failed to verify digest again.

martyone: OK, now I see. This is because the only change to the original target is that the local repo has been refreshed, which is not enough to imply snapshot reset. I would consider an alternative approach that would 1) be closer to the source of the issue and also 2) cover other mb2 invocations possibly appearing in future without the need to keep this behavior in mind.

So, could you please try whether the following change alone (without the `build-requires reset`) fixes it? (around line 351)

```diff
+    # Remove previously added repository to ensure that the SSU config changes.
+    # Otherwise the zypp cache update would not be propagated to the build target snapshots.
+    sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -R -m sdk-install ssu rr local-$DEVICE-hal >/dev/null 2>&1
     sb2 -t $VENDOR-$DEVICE-$PORT_ARCH -R -m sdk-install ssu ar local-$DEVICE-hal file://$LOCAL_REPO >>$LOG 2>&1|| die "can't add repo to target"
```

sledges: Yes, it worked!

martyone (approved):

martyone (approved):

sledges:
Rebased after merging `--package-timeline` PR.

PR #323, opened 2022-04-05
dhd: Immediately fail Android builds on OBS if a command fails

Thaodan:
Run the Android build commands with set -e/errexit so that in case on
them fails the shell directly exits instead of silently continuing.

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -398,9 +398,16 @@ else
     echo "Kernel Makefile not found, skipping EXTRAVERSION appending"
 fi
 
-# Hadk style build of android on OBS 
+# Hadk style build of android on OBS
+
+# We can check if we have new or old ubu-chroot by checking if it has the -V  option
+# added with this version.
+if ubu-chroot -V ; then
+   sh="sh -c"
+fi
+
 echo Running droid build in HABUILD_SDK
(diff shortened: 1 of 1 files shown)


sledges:
Old `ubu-chroot` (before https://github.com/sailfishos/android-tools-hadk/pull/7) will work incorrectly on OBS when invoked with `sh -c`.
Could you provide a safe-guard check of which `ubu-chroot` is being used?

sledges:
Nitpick: typo: to -> too.

Please fall back to the old cmdline so the transition is less painful for everybody:)

Thaodan:
> Nitpick: typo: to -> too.
> 
> Please fall back to the old cmdline so the transition is less painful for everybody:)

Adjusted please review.

droid-hal-device.inc
```diff
@@ -399,8 +399,12 @@ else
 fi
 
 # Hadk style build of android on OBS 
+if ubu-chroot -v ; then
+   sh="sh -c"
```
sledges, line 403: LGTM, worth commenting that argument parsing has been introduced together with version reporting.

sledges:
nit: obs->OBS (commit, body, and PR desc)

sledges:
in case on them fails->
in case one of them fails

sledges:
> in case on them fails-> in case one of them fails

one them -> one of them

sledges (changes requested):
Thanks for your patience:)

PR #324, opened 2022-04-25
helpers: Share sb2 target per family instead of per device and allow overriding scratchbox2 target used during build

Thaodan:
- helpers: Share sb2 target per family instead of per device
- helpers: Allow overriding of scratchbox2 target used during build


helpers/util.sh
```diff
@@ -325,17 +329,18 @@ build() {
...
     for SPEC in "$@" ; do
         minfo "Building $SPEC"
         mb2 \
             -s $SPEC \
-            -t $VENDOR-$DEVICE-$PORT_ARCH \
+            -t $HABUILD_TARGET \
             $PACKAGE_TIMELINE $NO_AUTO_VERSION \
             --output-dir "$LOCAL_REPO" \
-            build >>$LOG 2>&1|| die "building of package failed"
+            build --no-check >>$LOG 2>&1|| die "building of package failed"
```
sledges, line 343: I've built packages for `i3113`, then ran `export DEVICE=i4113` and mb2 is still checking places of the previous device:
```
PlatformSDK $ANDROID_ROOT $ rpm/dhd/helpers/build_packages.sh --version
* Building rpm/droid-hal-version-i4113.spec 
Checking whether to refresh metadata for dir:$ANDROID_ROOT/hybris/droid-hal-version-i3113/.mb2/filtered-output-dir
```

helpers/util.sh
```diff
@@ -71,7 +71,7 @@ mkdir -p $LOCAL_REPO
 PLUS_LOCAL_REPO="--plus-repo $LOCAL_REPO"
```
sledges, line 71: Two lines above it sets `LOCAL_REPO=$ANDROID_ROOT/droid-local-repo/$DEVICE` so even when using target for common family of boards, middleware RPMs would still need to somehow be made available for each $DEVICE in its local repo (TMTOWTDI).

sledges (changes requested):
We have [renamed](https://paste.opensuse.org/99654bb8) the terms to "build target" instead of "sb2 target", please revise your commit descriptions.

Minor typos:
would be than build per device ->
would be then build per device

SailfishOS -> Sailfish OS

-- 
This PR lays the groundwork for target override to be able to support multi-device multi-version build environments in the future.

helpers/util.sh
Thaodan: Hm for kumano this works fine for me: 
```
bidar@odin:~/dev/sailfish/hadk/kumano|21:59$ ./rpm/dhd/helpers/build_packages.sh --version
* Building rpm/droid-hal-version-j9210.spec 
Checking whether to refresh metadata for dir:/home/bidar/dev/sailfish/hadk/kumano/hybris/droid-hal-version-sony-kumano/.mb2/filtered-output-dir
```

helpers/util.sh
Thaodan: I changed the local repo to be shared too.

helpers/util.sh
```diff
@@ -70,6 +70,10 @@ LOCAL_REPO=$ANDROID_ROOT/droid-local-repo/$DEVICE
 mkdir -p $LOCAL_REPO
 PLUS_LOCAL_REPO="--plus-repo $LOCAL_REPO"
 
+if [ -z $HABUILD_TARGET ] ; then
```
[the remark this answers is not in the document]
Thaodan: Should we check for that or ask the to take care of this? The SFOS version isn't directly an indicator of breaking changes, it can but it doesn't have too.
Often breaking changes are in this repo or droid-configs and its dependencies.

PR #325, opened 2022-04-29
helpers: Remove revision number check from ssu-kickstarts.

mkosola:


Thaodan:
Aren't you removing the revision number rather than relaxing it?

mkosola:
> Aren't you removing the revision number rather than relaxing it?

Adjusted commit message

Thaodan:
Revision not version.

mkosola:
> Revision not version.

done

Thaodan (approved):

PR #328, opened 2022-07-03
Fix for hybris-18.1

foxclore:
This prevents kernel check on config in DTB*_OBJ


droid-hal-device.inc
```diff
@@ -417,7 +417,7 @@ mkdir tmp
 echo Verifying kernel config
 # AOSP seems to use .../obj/kernel/.config not obj/KERNEL_OBJ/.config like CM : so wildcard it
 hybris/mer-kernel-check/mer_verify_kernel_config \
-    %{android_root}/out/target/product/%{device}/obj/*/.config
+    %{android_root}/out/target/product/%{device}/obj/KERNEL_OBJ/.config
```
Thaodan, line 420: The read the comment above ^ you hard coding the path breaks anything else than anything based on CM.

Thaodan (changes requested):
Read the com

PR #329, opened 2022-08-25
dhd: Do not install /sbin/mkdosfs binary.

mkosola:
[dhd] Do not install /sbin/mkdosfs binary. JB#58522

Some Android build compile /sbin/mkdosfs binary which is
conflicting to busybox-symlinks-dosfstools in image building phase.

Signed-off-by: Matti Kosola <matti.kosola@jolla.com>

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -776,7 +776,8 @@ _remove_cruft   "/fstab.*"          \
                 "/charger"          \
                 "/res"              \
                 "/data"             \
-                "/odm/*"
+                "/odm/*"            \
+                "/sbin/mkdosfs"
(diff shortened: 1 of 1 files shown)


droid-hal-device.inc
```diff
@@ -776,7 +776,8 @@ _remove_cruft   "/fstab.*"          \
                 "/charger"          \
                 "/res"              \
                 "/data"             \
-                "/odm/*"
+                "/odm/*"            \
+                "/sbin/mkdosfs"
```
Thaodan, line 780: Would it make sense to exclude the whole path?

mkosola: Not sure if we can exclude everything from sbin, maybe leave it for later cleanup.

PR #335, opened 2023-06-10
Fixes for building with sony_aosp_11

moralrecordings:
- By default, Sony's msm-4.14 kernel will output temporary files to `out/clang.sh/kernel-tmp-(device)`. These are required by `rpm/dhd/helpers/build_packages.sh --droid-hal`. As the existing wildcard couldn't be modified easily to include such a different location, I replaced it with some conditionals and an exit statement on failure. The code now stores it in a shell variable KERNEL_PATH; I wasn't able to find a way to persist this between `%build` and `%install`, so I had to duplicate it. I'm not an RPM expert, so I'd welcome any advice.
- At some stage, the scripts were refactored to use zypper's `--plus-repo` option, which re-reads the repo before doing each operation. This doesn't work if GPG signing is not set up, leading to all zypper calls failing with `Warning: File 'repomd.xml' from repository 'dir:/[repo_path]' is unsigned.`. Fix is to add `--no-gpgcheck` to the local repo arguments.


droid-hal-device.inc
```diff
@@ -414,11 +414,22 @@ ubu-chroot -r /srv/mer/sdks/ubu ${bash} "set -o errexit; cd %android_root; %{?pr
...
 mkdir tmp
 
-echo Verifying kernel config
-# AOSP seems to use .../obj/kernel/.config not obj/KERNEL_OBJ/.config like CM : so wildcard it
-hybris/mer-kernel-check/mer_verify_kernel_config \
-    %{android_root}/out/target/product/%{device}/obj/*/.config
+# Find the kernel build directory.
+# AOSP seems to use .../obj/kernel/.config not obj/KERNEL_OBJ/.config like CM.
+# Sony AOSP 11 uses a separate storage area.
+if [[ -f %{android_root}/out/target/product/%{device}/obj/KERNEL_OBJ/.config ]]; then
```
Thaodan, line 420: Use [ instead.

droid-hal-device.inc
```diff
@@ -414,11 +414,22 @@ ubu-chroot -r /srv/mer/sdks/ubu ${bash} "set -o errexit; cd %android_root; %{?pr
...
+    KERNEL_PATH=%{android_root}/out/target/product/%{device}/obj/KERNEL_OBJ/
```
Thaodan, line 421: Don't caps the variable unless it is a environment variable.

droid-hal-device.inc
```diff
@@ -535,6 +546,20 @@ android_version_major=$(awk '/PLATFORM_VERSION([A-Z0-9.]*|_LAST_STABLE) := ([0-9
                           < %{android_root}/build/core/version_defaults.mk \
                           | awk -F'.' '{print $1}')
 
+# Find the kernel build directory.
+# AOSP seems to use .../obj/kernel/.config not obj/KERNEL_OBJ/.config like CM.
+# Sony AOSP 11 uses a separate storage area.
```
Thaodan, line 551: There isn't really a standard for these as you are not supposed to be build the kernel inside the Android build system.

Also the comment about CM is out of date as there's no CM anymore

Thaodan (commented):
I'm not sure if it is good to look for the path of the kernel config this way.
Something else would work better.

Thaodan:
@moralrecordings Please try #336 instead.

moralrecordings:
#336 is definitely the better approach, but is missing the check for `%{android_root}/out/clang.sh/kernel-tmp-%{device}/` as a potential path. 

The cheap option would be to move the search up to `%{android_root}/out`, which gives a 24 second wait on a hard disk. As you pointed out, we're rummaging around in the build artifacts so there isn't a standard location for anything.

Thaodan:
> #336 is definitely the better approach, but is missing the check for `%{android_root}/out/clang.sh/kernel-tmp-%{device}/` as a potential path.

That's the default path you should use -O <dir> to the regular path or create a PR to change the default.

PR #343, opened 2024-09-12
dhd: Allow having custom vendor_boot image

mlehtima:
[dhd] Allow having custom vendor_boot image

### DIFF
--- a/droid-hal-device.inc
+++ b/droid-hal-device.inc
@@ -58,6 +58,9 @@
 # have_custom_img_recovery:
 #                    defined when img-recovery is packaged separately
 #                      from droid-hal-device
+# have_custom_vendor_boot:
+#                    defined when vendor_boot is packaged separately from
+#                      droid-hal-device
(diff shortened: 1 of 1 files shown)


droid-hal-device.inc
```diff
@@ -1176,8 +1203,13 @@ done
...
 
+%if 0%{!?have_custom_vendor_boot:1}
 %if 0%{?enable_vendor_boot_update:1}
 %files vendor_boot
 /boot/vendor_boot.img
 %defattr(644,root,root,-)
 %endif
+%else
+%files kernel-modules-vendor_boot -f kernel-modules-vendor_boot.files
+%defattr(644,root,root,-)
```
Thaodan, line 1214: Do we need defattr here?

Thaodan (approved):
LGTM except nitpick comment.

droid-hal-device.inc
mlehtima: I'll do a thorough cleanup of the spec some other time.

